PR #1, opened 2016-07-01
Fix undefined/null mixup in documentation.

DieterDePaepe:


RubenVerborgh:
Thanks, `undefined` was indeed the old value (but that was incompatible with Node.js streams).

Issue #2, opened 2018-02-23
Transform functions do not maintain autoStart flag


rubensworks:
Consider the following example:
```
const AsyncIterator = require('AsyncIterator');

const it = new AsyncIterator.BufferedIterator({ autoStart: false });
let counter = 0;
it._read = function(count, cb) {
  console.log('Reading ' + count);
  for (let i = 0; i < count; i++) {
    this._push(counter++);
  }
  cb();
};

console.log('Initialized buffered iterator!');

const it2 = it.map((i) => i * 10);

console.log('Initialized mapped iterator!');
```

As the `autoStart` parameter of `it` is set to `false`, no elements will be buffered until a data listener is attached.
However, when the iterator is transformed (for example using `map`), then this `autoStart` parameter is not inherited.

For this example, the following output will be produced:
```
Initialized buffered iterator!
Initialized mapped iterator!
Reading 4
Reading 1
Reading 1
Reading 1
Reading 1
```

The expected behaviour would be that no elements are produced until a data listener is attached to either `it` or `it2`.

RubenVerborgh:
I've been thinking a bit about this, and I'm not sure what the desired behavior is:

1. On the one hand, one could argue that, since the source stream is started lazily, so should the transformed stream.

2. On the other hand, one could argue that, by calling `transform` on the source stream, you are essentially starting it.

I started thinking that this second behavior is actually more natural.
So in cases where you want the first behavior, you would need to pass the `autoStart: false` attribute to the `.transform` function (which isn't available through shorthands such as `map`).

Given the above arguments, what do you think?
I've created a branch `fix-transform-autostart` which has `autoStart: false` behavior as the default for `transform` and its helpers so you can compare. (That said, it's the only place where `autoStart` would not be true by default.) The exact behavior you suggest here, having `autoStart` equal to the source, could also be implemented, but would complicate matters (so I first want to have the discussion about its necessity).

-----

Minor nit:

> The expected behaviour would be that no elements are produced until a data listener is attached to either it or it2.

You shouldn't read `it` after you have mapped it (as stated in the docs).

rubensworks:
I think both options make sense.

[...]

PR #3, opened 2018-02-25
Group sync transformations in same event loop run

RubenVerborgh:
In #2, I noticed that the `map` operation caused 4 reads of 1 item, instead of 1 read of 4 items. Turns out that `SimpleTransformIterator` awaits the next tick, even if no asynchronous transformation is performed.

This pull request ensures that synchronous transformations are executed directly after one another when possible.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -137,7 +137,7 @@
     new-parens: 2,
     newline-after-var: 0,
     no-array-constructor: 2,
-    no-continue: 2,
+    no-continue: 0,
     no-inline-comments: 0,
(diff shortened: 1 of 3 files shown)


RubenVerborgh:
@rubensworks Your report for #2 revealed a performance for synchronous transformations. Could you test this patched version with Comunica to see if you don't run into any issues?

rubensworks:
Just tested this patch and everything (unit tests + queries) still seems to work as intended.

RubenVerborgh:
Thanks, merged!

Issue #6, opened 2018-06-21
Closing/destroying sources when done


rubensworks:
In https://github.com/comunica/comunica/issues/144, we discussed adding a `closeSourceWhenDone` (_defaults to `true`_) to `TransformIterator `. This would make the transformed iterator call `close()` on its source when the transformed iterator has ended.

While this would work in practise, it may not be the cleanest solution, as `Readable#close` is not part of the [official interface](https://nodejs.org/api/stream.html#stream_readable_destroy_error).
Instead, [`Readable#destroy`](https://nodejs.org/api/stream.html#stream_readable_destroy_error) _is_ part of the interface, and can be used to achieve the same thing (and has been [available for a long time](https://github.com/nodejs/node-v0.x-archive/blob/cfcb1de130867197cbc9c6012b7e84e08e53d032/lib/fs.js#L1587-L1594)).

So I'm wondering if it would be better to instead use a property `destroySourceWhenDone` that calls `destroy()` on the source. This would probably require `AsyncIterator` to implement the `destroy()` method as well, which could simply be a proxy for `close()`.

RubenVerborgh:
It seems that `destroy` has stronger semantics than `close`.

### destroy

> Destroy the stream, and emit 'error' and 'close'. After this call, the readable stream will release any internal resources and subsequent calls to push() will be ignored. Implementors should not override this method, but instead implement readable._destroy().

### close

> Stops the iterator from generating new items.
Already generated items or terminating items can still be emitted. After this, the iterator will end asynchronously.

Our options:
- implement an actual `destroy` on `AsyncIterator` (no real necessity, and can have some complexities)
- assume the existence of `close`
- try `close`
- try `close`, fall back to `destroy`

rubensworks:
> implement an actual destroy on AsyncIterator (no real necessity, and can have some complexities)

I think this first option would be the most robust, and it strictly uses defined methods.

> try close, fall back to destroy

This last one might also be a good possibility, but I'm afraid that this may result in unexpected situations depending on the type of source streams.

In any case, it's your library, so it's your call in any case 😉

RubenVerborgh:
>> implement an actual destroy on AsyncIterator (no real necessity, and can have some complexities)
> 
> I think this first option would be the most robust, and it strictly uses defined methods.

[...]

PR #7, opened 2018-06-27
Feature/destroy

rubensworks:
As discussed in #6, this adds a `destroy` function that is compatible with Node's ReadableStream API.

This introduces a new `DESTROYED`, a `destroyed` getter and a `done` (destroyed or ended).

Breaking change:
* Calling `_push` after being ended will not throw an error anymore.

(This does not yet include automatically destroying transformed sources when done, this will be added in a separate PR)

Be sure to double-check the unit tests, they *should* cover all cases.

### DIFF
--- a/asynciterator.js
+++ b/asynciterator.js
@@ -606,7 +606,6 @@ ArrayIterator.prototype._toStringDetails = function () {
 
 /* Called by {@link AsyncIterator#destroy} */
 ArrayIterator.prototype._destroy = function (error, callback) {
-  // Release our complete buffer.
   this._buffer = null;
   callback();
 };
@@ -934,7 +933,6 @@ BufferedIterator.prototype._completeClose = function () {
 
 /* Called by {@link AsyncIterator#destroy} */
 BufferedIterator.prototype._destroy = function (error, callback) {
-  // Release our complete buffer.
   this._buffer = [];
   callback();
 };
--- a/asynciterator.js
+++ b/asynciterator.js
@@ -1092,7 +1092,7 @@ TransformIterator.prototype._closeWhenDone = function () {
 };
 
 /* Cleans up the source iterator and ends. */
-TransformIterator.prototype._end = function () {
+TransformIterator.prototype._end = function (destroy) {
   var source = this._source;
   if (source) {
     source.removeListener('end',      destinationCloseWhenDone);
@@ -1100,7 +1100,7 @@ TransformIterator.prototype._end = function () {
     source.removeListener('readable', destinationFillBuffer);
     delete source._destination;
   }
-  BufferedIterator.prototype._end.call(this);
+  BufferedIterator.prototype._end.call(this, destroy);
 };
 
 /**
@@ -1633,14 +1633,14 @@ ClonedIterator.prototype.read = function () {
 };
 
 /* End the iterator and cleans up. */
-ClonedIterator.prototype._end = function () {
+ClonedIterator.prototype._end = function (destroy) {
   // Unregister from a possible history reader
   var history = this._source && this._source._destination;
   if (history) history.unregister(this);
 
   // Don't call TransformIterator#_end,
   // as it would make the source inaccessible for other clones
-  BufferedIterator.prototype._end.call(this);
+  BufferedIterator.prototype._end.call(this, destroy);
 };
 
 // Disable buffer cleanup
--- a/asynciterator.js
+++ b/asynciterator.js
@@ -826,11 +826,11 @@ BufferedIterator.prototype._read = function (count, done) { done(); };
   @emits AsyncIterator.readable
 **/
 BufferedIterator.prototype._push = function (item) {
-  if (this.done)
-    throw new Error('Cannot push after the iterator was ended.');
-  this._pushedCount++;
-  this._buffer.push(item);
-  this.readable = true;
+  if (!this.done) {
+    this._pushedCount++;
+    this._buffer.push(item);
+    this.readable = true;
+  }
 };
 
 /**
--- a/asynciterator.js
+++ b/asynciterator.js
@@ -24,8 +24,8 @@ var immediate = (function () {
   @type String[]
   @protected
 */
-var STATES = AsyncIterator.STATES = ['INIT', 'OPEN', 'CLOSING', 'CLOSED', 'ENDED'];
-var INIT = 0, OPEN = 1, CLOSING = 2, CLOSED = 3, ENDED = 4;
+var STATES = AsyncIterator.STATES = ['INIT', 'OPEN', 'CLOSING', 'CLOSED', 'ENDED', 'DESTROYED'];
+var INIT = 0, OPEN = 1, CLOSING = 2, CLOSED = 3, ENDED = 4, DESTROYED = 5;
 STATES.forEach(function (state, id) { AsyncIterator[state] = id; });
 
 /**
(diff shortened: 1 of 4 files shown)


asynciterator.js
```diff
@@ -209,24 +220,63 @@ AsyncIterator.prototype.close = function () {
...
+};
+
+/**
+ Called by {@link AsyncIterator#destroy}.
+ Implementors can override this, but this should not be called directly.
+
+ @param {Error} error A possible error.
+ @param {Function} callback A callback function with an optional error argument.
+ */
+AsyncIterator.prototype._destroy = function (error, callback) {
```
RubenVerborgh: Why would this one take an error as argument?

asynciterator.js
```diff
@@ -209,24 +220,63 @@ AsyncIterator.prototype.close = function () {
...
+ even not already generated items.
+
+ Implementors should not override this method,
+ but instead implement {@link AsyncIterator#_destroy}.
+
+ @param {Error} [error] An optional error to emit.
+ @emits AsyncIterator.end
+ @emits AsyncIterator.error Only emitted if an error is passed.
+ **/
+AsyncIterator.prototype.destroy = function (error) {
```
RubenVerborgh: Maybe call this one `cause`.

asynciterator.js
```diff
@@ -209,24 +220,63 @@ AsyncIterator.prototype.close = function () {
...
   typically, `close` is responsible for calling `_end`.
 
+  @param {boolean} [destroy] If the iterator should be forcefully destroyed.
   @protected
   @emits AsyncIterator.end
 **/
-AsyncIterator.prototype._end = function () {
-  if (this._changeState(ENDED)) {
+AsyncIterator.prototype._end = function (destroy) {
+  if (this._changeState(destroy ? DESTROYED : ENDED)) {
```
RubenVerborgh, line 272: Hmm, we should do a careful check of all places that can call `_end`.

Because `changeState` guarantees that we cannot go from a higher state to a lower state, but we could still have the case where `_end` is called on an iterator in state `ENDED`, which would transition it to `DESTROYED`.

Alternatively, we can change `changeState` such that it can never transition beyond `ENDED`.

asynciterator.js
```diff
@@ -317,7 +392,7 @@ function emitData() {
   while (this._hasListeners('data') && (item = this.read()) !== null)
     this.emit('data', item);
   // Stop draining the source if there are no more `data` listeners
-  if (!this._hasListeners('data')) {
+  if (!this._hasListeners('data') && !this.done) {
```
RubenVerborgh, line 395: What condition do we catch with this? Was this a bug that already existed, or a new condition introduced because of destroyed?

asynciterator.js
```diff
@@ -529,6 +604,13 @@ ArrayIterator.prototype._toStringDetails = function () {
   return '(' + (this._buffer && this._buffer.length || 0) + ')';
 };
 
+/* Called by {@link AsyncIterator#destroy} */
+ArrayIterator.prototype._destroy = function (error, callback) {
+  // Release our complete buffer.
```
RubenVerborgh: Comment can go.

asynciterator.js
```diff
@@ -850,6 +932,13 @@ BufferedIterator.prototype._completeClose = function () {
   }
 };
 
+/* Called by {@link AsyncIterator#destroy} */
+BufferedIterator.prototype._destroy = function (error, callback) {
+  // Release our complete buffer.
```
RubenVerborgh: Comment can go.

asynciterator.js
```diff
@@ -529,6 +604,13 @@ ArrayIterator.prototype._toStringDetails = function () {
   return '(' + (this._buffer && this._buffer.length || 0) + ')';
 };
 
+/* Called by {@link AsyncIterator#destroy} */
+ArrayIterator.prototype._destroy = function (error, callback) {
+  // Release our complete buffer.
+  this._buffer = null;
```
RubenVerborgh: Or empty array? (Might be faster since then there's only one internal V8 class). However, I do see `this._buffer` checks above, so might be null in other cases.

test/AsyncIterator-test.js
```diff
@@ -367,6 +624,82 @@ describe('AsyncIterator', function () {
     });
   });
 
+  describe('An AsyncIterator instance to which 2 items are added an will be destroyed', function () {
+    var iterator, items = [], dataListener1;
```
RubenVerborgh: Why the `1`?

test/AsyncIterator-test.js
```diff
@@ -367,6 +624,82 @@ describe('AsyncIterator', function () {
...
+      });
+
+      it('should not be listening for the `newListener` event', function () {
+        EventEmitter.listenerCount(iterator, 'newListener').should.equal(0);
+      });
+    });
+  });
+
+  describe('An AsyncIterator instance to which 2 items are added an will be destroyed with an error', function () {
+    var iterator, items = [], err, dataListener1, errorListener1;
```
RubenVerborgh: Why the `1`?

test/BufferedIterator-test.js
```diff
@@ -1843,4 +2411,114 @@ describe('BufferedIterator', function () {
...
+
+  describe('A BufferedIterator that is destroyed after the first item', function () {
+    var iterator, i = 0;
+    before(function () {
+      iterator = new BufferedIterator();
+      iterator._read = sinon.spy(function (count, next) {
+        this._push(++i);
+        next();
+        if (i === 1)
+          iterator.destroy();
```
RubenVerborgh, line 2480: Worth testing the case where this happens before `next`?

asynciterator.js
rubensworks: This is the error that was passed to `destroy`. This may be useful for implementors of `_destroy` to know if the stream was made to error or not.

This is how it's done in Node's `ReadableStream` was well.

asynciterator.js
rubensworks: `destroy` has a `if (!this.done)` check, so it won't do anything if the asynciterator is in the `ENDED` state.

I explicitly added a test where `destroy` is called after `close` here: https://github.com/rubensworks/AsyncIterator/blob/d2e1114be1ec2a4dd5fe92146e9c4eac0aa41010/test/AsyncIterator-test.js#L150-L191

Do you think there are other cases that need to be checked/tested as well?

asynciterator.js
rubensworks: Without this check, the _"should not be listening for the `newListener` event"_ tests fail after calling `destroy`. This is because forcefully destroying the stream can still have generating items in progress, and this waiting-loop will therefore be still running. The way this waiting-loop is implemented at the moment, I don't think there's another way to handle this.

asynciterator.js
rubensworks: There's a `delete this._buffer` call above, which is better indeed. I'll do the same here.

rubensworks:
Resolved all comments, expect for the `_destroy`, `_end` and `emitData` issues.

asynciterator.js
RubenVerborgh: Better to be safe than sorry. I made the state transition from `ENDED` to `DESTROYED` impossible in dabb6844e891786e1c319ad574c7103c4e54f2f6.

asynciterator.js
```diff
@@ -1633,14 +1633,14 @@ ClonedIterator.prototype.read = function () {
...
-ClonedIterator.prototype._end = function () {
+ClonedIterator.prototype._end = function (destroy) {
   // Unregister from a possible history reader
   var history = this._source && this._source._destination;
   if (history) history.unregister(this);
 
   // Don't call TransformIterator#_end,
   // as it would make the source inaccessible for other clones
-  BufferedIterator.prototype._end.call(this);
+  BufferedIterator.prototype._end.call(this, destroy);
```
RubenVerborgh, line 1643: @rubensworks :wink:

RubenVerborgh:
@rubensworks Merged to develop. Can you verify this works well with comunica/comunica#144? If so, I will release v2.0.0.

rubensworks:
Yep, I'll first make a PR for the auto-closing.

PR #8, opened 2018-07-05
Add `destroySourceWhenDone` option to TransformIterator

rubensworks:
This value defaults to true, and ensures that when the transformed
iterator is ended, the source iterator is destroyed.

### DIFF
--- a/test/TransformIterator-test.js
+++ b/test/TransformIterator-test.js
@@ -611,6 +611,46 @@ describe('TransformIterator', function () {
     });
   });
 
+  describe('A TransformIterator with destroySource set to its default', function () {
+    var iterator, source;
+    before(function () {
+      source = new ArrayIterator([1, 2, 3]);
+      iterator = new TransformIterator(source);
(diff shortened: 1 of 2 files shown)


test/TransformIterator-test.js
```diff
@@ -215,6 +215,7 @@ describe('TransformIterator', function () {
     before(function () {
       iterator = new TransformIterator(source = new AsyncIterator());
       captureEvents(iterator, 'readable', 'end');
+      captureEvents(source, 'close');
```
RubenVerborgh: What purpose does this serve? AsyncIterator never emits close.

RubenVerborgh:
@rubensworks Amended some changes, in particular:
- renaming the option to the shorter `destroySource`
- adding unit tests that check destruction in both cases

I also removed a `close` test that I found strange.

Can you verify whether everything works as intended?

rubensworks:
Everything still works :-)

PR #15, opened 2020-06-19
3.x.x enhancements

rubensworks:
Several minor enchancements related to the 3.x.x update.

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -491,7 +491,7 @@ export class AsyncIterator<T> extends EventEmitter {
     After this operation, only read the returned copies instead of the original iterator.
     @returns {module:asynciterator.AsyncIterator} A new iterator that contains all future items of this iterator
   */
-  clone(): AsyncIterator<T> {
+  clone(): ClonedIterator<T> {
     return new ClonedIterator<T>(this);
   }
 }
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -968,6 +968,7 @@ export class TransformIterator<S, D = S> extends BufferedIterator<D> {
   protected _sourcePromise?: Promise<Source<S>>;
   protected _destroySource: boolean;
   protected _optional: boolean;
+  protected _boundPush = (item: D) => this._push(item);
 
   /**
     Creates a new `TransformIterator`.
@@ -982,11 +983,11 @@ export class TransformIterator<S, D = S> extends BufferedIterator<D> {
   constructor(source?: AsyncIterator<S> | Promise<AsyncIterator<S>>,
               options: TransformIteratorOptions<S> =
                 source as TransformIteratorOptions<S> || {}) {
+    super(options);
+
     // Shift parameters if needed
     if (!source || !(isEventEmitter(source) || isPromise(source)))
       source = options.source;
-    super(options);
-
     // The passed source is an AsyncIterator or readable stream
     if (isEventEmitter(source)) {
       this.source = source;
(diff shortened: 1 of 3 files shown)


asynciterator.ts
```diff
@@ -491,7 +491,7 @@ export class AsyncIterator<T> extends EventEmitter {
     After this operation, only read the returned copies instead of the original iterator.
     @returns {module:asynciterator.AsyncIterator} A new iterator that contains all future items of this iterator
   */
-  clone(): AsyncIterator<T> {
+  clone(): ClonedIterator<T> {
```
RubenVerborgh, line 494: Is there a reason we want to give the more specific class? I thought to keep `ClonedIterator` (more or less) internal.

asynciterator.ts
```diff
@@ -1084,7 +1084,7 @@ export class TransformIterator<S, D = S> extends BufferedIterator<D> {
     if (!source || source.done || (item = source.read()) === null)
       done();
     else if (!this._optional)
-      this._transform(item, next);
+      this._transform(item, next, this._push.bind(this));
```
RubenVerborgh: probably want a stored `item => this._push(item)` for performance (I can update)
To basically, in the constructor/member variable declaration, `this._pushBound = item => this._push(item)`.

asynciterator.ts
```diff
@@ -1099,7 +1099,7 @@ export class TransformIterator<S, D = S> extends BufferedIterator<D> {
       if (pushedCount === this._pushedCount)
         this._push(item as any as D);
       done();
-    });
+    }, this._push.bind(this));
```
RubenVerborgh: idem

asynciterator.ts
```diff
@@ -1252,7 +1253,7 @@ export class SimpleTransformIterator<S, D = S> extends TransformIterator<S, D> {
       // Asynchronously transform the item, and wait for `next` to call back
       else {
         if (!this._optional)
-          this._transform(mappedItem as S, next);
+          this._transform(mappedItem as S, next, this._push.bind(this));
```
RubenVerborgh: idem

asynciterator.ts
```diff
@@ -1304,6 +1305,36 @@ export class SimpleTransformIterator<S, D = S> extends TransformIterator<S, D> {
...
+   @param {Array|module:asynciterator.AsyncIterator} [options.prepend] Items to insert before the source items
+   @param {Array|module:asynciterator.AsyncIterator} [options.append]  Items to insert after the source items
+   */
+  constructor(source: AsyncIterator<S>,
+              options?: MultiTransformOptions<S, D> |
+                        MultiTransformOptions<S, D> & ((item: S) => AsyncIterator<D>)) {
+    super(source, options as TransformIteratorOptions<S>);
+
+    // Set transformation steps from the options
+    if (options) {
```
RubenVerborgh, line 1332: This is just new functionality? Or something that broke? (Going to keep it in any case.)

asynciterator.ts
rubensworks: I just checked again, and it was not really needed.

But I would still consider keeping the more specific type though, as ClonedIterator (and TransformIterator) will have some more typed elements, such as `.source`.
But no strong opinion.

asynciterator.ts
rubensworks: +1 to storing. I wasn't aware that `.bind` is less performant than a lambda.

asynciterator.ts
rubensworks: Yep, that's just new. I found it annoying that I had to create dedicated classes just for doing small multi-transformations.

asynciterator.ts
RubenVerborgh: 👍 will keep

asynciterator.ts
RubenVerborgh: 👍 will keep

PR #23, opened 2020-09-23
Ensure transformations without autoStart are lazy

### LINKED ISSUE #22: Chained transformations make empty iterators un-lazy
When transforming an empty lazy stream, the final stream becomes un-lazy.
This makes it so that the end-event can be emitted before any listeners can be attached.

Minimal reproducible example:
```javascript
(async function() {
  // Create a lazy array it
  const arrayIt = new ArrayIterator([], { autoStart: false });

  // Transform it twice (both lazy)
  const transformedIt1 = arrayIt.transform({
    map: () => 1,
    autoStart: false,
  });
  const transformedIt2 = transformedIt1.transform({
    map: () => 1,
    autoStart: false,
  });

  console.log('WAITING...');
  setTimeout(() => {
    console.log('ATTACHING LISTENERS...');

    // The listeners below are never called
    transformedIt2.on('data', () => console.log('DATA'));
    transformedIt2.on('end', () => console.log('ENDED'));
  }, 1000);
})();
```

RubenVerborgh:
This PR adds an internal flag to indicate whether a source has been started.

Fixes https://github.com/RubenVerborgh/AsyncIterator/issues/22

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -986,6 +986,7 @@ export class BufferedIterator<T> extends AsyncIterator<T> {
 export class TransformIterator<S, D = S> extends BufferedIterator<D> {
   protected _source?: InternalSource<S>;
   protected _createSource?: (() => AsyncIteratorOrPromise<S>) | null;
+  protected _sourceStarted: boolean;
   protected _destroySource: boolean;
   protected _optional: boolean;
   protected _boundPush = (item: D) => this._push(item);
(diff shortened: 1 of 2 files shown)


RubenVerborgh:
I'm amazed that we're still reaching 61% with only 2 tests. Take note, @joachimvh.
That said, let me switch on all tests again.

rubensworks:
@RubenVerborgh This change seems to fix all remaining "hanging queries" in Comunica 🎉 
Once this has been released, I'll update to it in https://github.com/comunica/comunica/pull/733.

RubenVerborgh:
Released as v3.0.2, thanks for your help to get to this bug and a fix fast.
Feel free to re-ping me in that issue when ready.

PR #27, opened 2021-03-09
Fix ClonedIterator not respecting sources without autoStart, Closes #26

### LINKED ISSUE #26: Cloning not inheriting autoStart: false
When a lazy (empty) iterator is cloned, it seems to consume the iterator eagerly, which makes it so that `'end'` events can be lost.

Minimal reproducible example:
```javascript
  const arrayIt = new ArrayIterator([], { autoStart: false });
  const clonedIt = arrayIt.clone();

  await new Promise((resolve) => setImmediate(resolve));

  // The listeners below are never called
  clonedIt.on('data', () => console.log('DATA'));
  clonedIt.on('end', () => console.log('ENDED'));
```

rubensworks:
This PR does the following:

* `ClonedIterator` does not eagerly pull the history if `source._sourceStarted` is `false`.
* The `_sourceStarted` flag is also added to `ArrayIterator`.
* The `_sourceStarted` flag is moved from `TransformIterator` to `BufferedIterator`, so that it also applies to the `UnionIterator`.

**Warning: this PR fixes the original problem, but I found a new edge-case that is not working yet.**
A (failing) integration test is included for this case.
I've spent quite a bit of time trying to find the problem, but the chained clones and history were breaking my head.
I hope to look into this later with a fresh mind, but if you have any thoughts/suggestions on where I could debug next, please let me know.

-----

I was wondering if it makes sense to move `_sourceStarted` up all the way to the base `AsyncIterator` class. But such a change may be too invasive, and may perhaps be done in a separate PR.

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -608,13 +608,11 @@ export class ArrayIterator<T> extends AsyncIterator<T> {
     super();
     const buffer = items ? [...items] : [];
     this._sourceStarted = autoStart !== false;
-    if (this._sourceStarted && buffer.length === 0) {
+    if (this._sourceStarted && buffer.length === 0)
       this.close();
-    }
-    else {
-      this.readable = true;
+    else
       this._buffer = buffer;
-    }
+    this.readable = true;
   }
 
   /* Reads an item from the iterator. */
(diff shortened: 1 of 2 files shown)


rubensworks:
@RubenVerborgh All issues have now been resolved.

The important elements in my latest changes are the following:
* `ArrayIterator` _always_ sets `readable` to true.
* `ClonedIterator` also sets `readable` to true when the source is lazy.
* The `HistoryReader` does not close clones that have not been started yet (if they are lazy).

All tests in Comunica pass with these changes, including the update-related tests that failed before.

RubenVerborgh:
Super, thank you so much! Quite the detective work.

Issue #28, opened 2021-06-17
Current task scheduler does not guarantee order


RubenVerborgh:
The snippet at https://github.com/RubenVerborgh/AsyncIterator/blob/668697c568484881f663b786e0b6a037c2cd5ab3/taskscheduler.ts#L13-L21 can result in out-of-order execution when a microtask is scheduled before `setTimeout` has occurred.

RubenVerborgh:
@rubensworks Could you test if https://github.com/RubenVerborgh/AsyncIterator/tree/fix/task-order does not break anything? This should be the final guard before #29 can be released. (It's weird that we have not run into out-of-order bugs yet.)

rubensworks:
Everything seems to work fine! (aside for #30 that is)

RubenVerborgh:
Thank you!

PR #29, opened 2021-06-24
Also alternate with macrotasks in Node.js

rubensworks:
As discovered in https://github.com/comunica/comunica/issues/826, some use cases may lead to extensive microtask usages, which even blocks the Node.js garbage collector, and leads to memory issues.

This PR modifies the micro/macrotasks alteration trick so that it also applies to Node.js by default.

(Fix tested in Comunica)


RubenVerborgh:
Looks good, but still concerned about https://github.com/RubenVerborgh/AsyncIterator/issues/28, which might become a bigger problem when we extend this behavior to all platforms.

RubenVerborgh:
In v3.2.0

PR #31, opened 2021-07-14
Feat/overload filter type

### LINKED ISSUE #30: [Feature] Overload filter type
I think it would be quite useful to have an overload on the filter method for the asyncIterator that is

```ts
filter<K extends T>(filter: (item: T) => item is K, self?: any): AsyncIterator<K>;
```

This would then allow use in cases such as the following:

```ts
const termIterator: AsyncIterator<RDF.Term> = /* Some iterator */
const namedNodeIterator: AsyncIterator<RDF.NamedNode> = termIterator.filter<RDF.NamedNode>((term): term is RDF.NamedNode => term.termType === 'NamedNode')
```

jeswr:
Closes #30


RubenVerborgh:
Thank you!

PR #32, opened 2021-07-19
Fix bug in https://github.com/RubenVerborgh/AsyncIterator/issues/30

jeswr:
Resolve bug with missing overload declaration.

I don't have time to test this now - I can create a PR that converts the test suite to TS and so can properly apply tests to this change if you would like me to do that first (I can do this within the next 24 hrs)

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -437,6 +437,7 @@ export class AsyncIterator<T> extends EventEmitter {
     @returns {module:asynciterator.AsyncIterator} A new iterator that filters items from this iterator
   */
   filter<K extends T>(filter: (item: T) => item is K, self?: any): AsyncIterator<K>;
+  filter(filter: (item: T) => boolean, self?: any): AsyncIterator<T>;
   filter(filter: (item: T) => boolean, self?: any): AsyncIterator<T> {
     return this.transform({ filter: self ? filter.bind(self) : filter });
   }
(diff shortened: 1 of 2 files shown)


RubenVerborgh:
@rubensworks ok?

rubensworks:
Yes, this fixes the problem. Thanks @jeswr!

jeswr:
@RubenVerborgh  I'm in the process of moving the test suite to TS as mentioned above. Is it ok if I migrate it to use Jest at the same time (it will make the process easier)?

RubenVerborgh:
@jeswr Be my guest but… I expect that to be a substantial amount of work. Especially with timing issues etc.

jeswr:
> I expect that to be a substantial amount of work. Especially with timing issues etc.

It is a bit, so it's a WIP here https://github.com/jeswr/AsyncIterator/tree/chore/tests-to-typescript.

In the meantime, I have added the `FilterTypeTest.ts` file which tests that both versions of the overload work (if there is a type error then there will be an error when the test suite is run), so this AFAIK this should be safe to merge.

RubenVerborgh:
Super, thank you!

RubenVerborgh:
This somehow removes it from the docs though: https://github.com/RubenVerborgh/AsyncIterator/commit/d3f24be62e16d26d6f4c445b46bb75fb3072882e

Issue #35, opened 2021-12-31
[Possible bug] Empty iterator does not work after `await`


jeswr:
The `empty` iterator does not emit events if something has been `await`ed after its instantiation - for instance, the following tests time out

```ts
async function promisifyEventEmitter(eventEmitter: EventEmitter): Promise<void> {
  return new Promise<void>((resolve, reject) => {
    eventEmitter.on('end', resolve);
    eventEmitter.on('error', reject);
  });
}

it('empty test awaiting promise', async () => {
  const e = empty();
  await Promise.resolve();
  await expect(promisifyEventEmitter(e)).resolves.toBeFalsy();
});

it('empty test undefined', async () => {
  const e = empty();
  await undefined;
  await expect(promisifyEventEmitter(e)).resolves.toBeFalsy();
});
```

whilst the following works

```ts
it('empty test undefined', async () => {
  const e = empty();
  await expect(promisifyEventEmitter(e)).resolves.toBeFalsy();
});
```

Discovered in https://github.com/comunica/comunica/issues/904

jeswr:
I've added some more tests with the behavior I would expect [here](https://github.com/jeswr/AsyncIterator/tree/fix/empty-iterator-await) - but haven't yet had the time to find the root cause of the issue (and not sure if I will anytime soon unfortunately).

All of the tests on this branch pass `npm run test:immediate`

```bash
node: v16.10.0
npm: 7.24.0
```

rubensworks:
I suspect you're running into an autoStart issue (see some of the closed issues here).
By default, asynciterators will start automatically after creation.
That means that if the iterator is empty, it will immediately end, so end-events may be emitted before listeners are attached.

There is an issue to change this behaviour: #25.

jeswr:
So in doing working on https://github.com/RubenVerborgh/AsyncIterator/pull/36 I've discovered that this issue is not entirely resolved by changing the `autoStart` to be false by default since the `EmptyIterator` does not actually have an `autoStart` parameter.

In particular the bottom two of these following tests fail

```ts

  describe('An EmptyIterator should not emit until read from', () => {
    it('no awaiting', async () => {
      const iterator = new EmptyIterator();
      await expect(await promisifyEventEmitter(iterator)).to.be.undefined;
    });

    it('awaiting undefined', async () => {
      const iterator = new EmptyIterator();
      await undefined;
      await expect(await promisifyEventEmitter(iterator)).to.be.undefined;
    });

[...]

PR #36, opened 2022-02-14
set autostart to false

jeswr:


asynciterator.ts
```diff
@@ -1220,6 +1220,18 @@ export class SimpleTransformIterator<S, D = S> extends TransformIterator<S, D> {
...
+          removeOptions = false;
+          break;
+        }
+      }
+      if (removeOptions)
+        options = undefined;
+    }
+
     // Set transformation steps from the options
     options = options || (!isSourceExpression(source) ? source : null as any);
```
jeswr, line 1236: This was the only cause of some problems in the testing - since passing `option = { autoStart: true }` means that the source is not used even if it is not a source expression.

I've got around this by setting the options to undefined if the only key is `autoStart` - but I'm thinking that the tests should just be changed instead

asynciterator.ts
```diff
@@ -1220,6 +1220,18 @@ export class SimpleTransformIterator<S, D = S> extends TransformIterator<S, D> {
                        TransformOptions<S, D> & ((item: S, done: () => void) => void)) {
     super(source, options as TransformIteratorOptions<S>);
 
+    if (typeof options === 'object') {
```
RubenVerborgh, line 1223: Could you enlighten me on this block? I'm not sure what this does and why.

test/ArrayIterator-test.js
```diff
@@ -43,10 +43,10 @@ describe('ArrayIterator', () => {
     });
   });
 
-  describe('An ArrayIterator without arguments', () => {
+  describe('An ArrayIterator undefined iterator and autostart', () => {
     let iterator;
     before(() => {
-      iterator = new ArrayIterator();
+      iterator = new ArrayIterator(undefined, { autoStart: true });
```
RubenVerborgh: That is a bit weird to me; I'll investigate.

test/BufferedIterator-test.js
```diff
@@ -30,7 +30,7 @@ describe('BufferedIterator', () => {
   describe('A BufferedIterator without arguments', () => {
     let iterator;
     before(() => {
-      iterator = new BufferedIterator();
+      iterator = new BufferedIterator({ autoStart: true });
```
RubenVerborgh: This is where this change will get a bit annoying, but I think it's important that the test structure remains the same. So first "without arguments" (so don't pass anything) and then the other cases.

(Note that the description and code are out of sync too.)

RubenVerborgh (commented):
Looking good; I think we'll get there!

asynciterator.ts
jeswr: This is to handle https://github.com/RubenVerborgh/AsyncIterator/pull/36#discussion_r805789594

test/ArrayIterator-test.js
jeswr: Was using undefined to emulate the existing test with no sources. An empty array should work as well instead.

asynciterator.ts
RubenVerborgh: I'm sorry, I've been looking at this for 15 minutes now but still don't understand.

So the logic of the `SimpleTransformIterator` is that both of its parameters (`source` and `options`) are optional.

So what the original line 1224 is doing is determining whether `options` is the first or the second argument (given that `source` is optional, it might change position). The easy case is when `options` is non-nullish: then `options` is the second argument. If `options` is nullish, then `options` is the first argument—if that first argument is not a source.

None of that should change, regardless of the `autoStart` default. So probably a test change indeed.

jeswr: Makes sense - will do a test change then :)

test/ArrayIterator-test.js
```diff
@@ -679,4 +595,88 @@ describe('ArrayIterator', () => {
       });
     });
   });
+
+  describe('An ArrayIterator undefined iterator and autostart', () => {
+    let iterator;
+    before(() => {
+      iterator = new ArrayIterator(undefined, { autoStart: true });
```
RubenVerborgh, line 602:
```suggestion
  describe('An ArrayIterator without source with autostart', () => {
    let iterator;
    before(() => {
      iterator = new ArrayIterator({ autoStart: true });
```

(should be made to work if it doesn't)

asynciterator.ts
```diff
@@ -1457,9 +1469,8 @@ export class UnionIterator<T> extends BufferedIterator<T> {
     @param {object} [options] Settings of the iterator
   */
   constructor(sources: AsyncIteratorOrArray<AsyncIterator<T>>,
-              options: BufferedIteratorOptions = {}) {
-    super(options);
-    const autoStart = options.autoStart !== false;
+              { autoStart = false, ...options }: BufferedIteratorOptions = {}) {
+    super({ autoStart, ...options });
```
RubenVerborgh, line 1473: I wouldn't deconstruct here; we can just use `options.autoStart` below. The original was needed because we didn't want to repeat the `!== false` cast, which is no longer needed (as in 6f3af334c76e792e674ffce3dd74b211ebb79308).

test/ArrayIterator-test.js
```diff
@@ -43,94 +43,10 @@ describe('ArrayIterator', () => {
...
-    it('should return null when read is called', () => {
-      expect(iterator.read()).to.be.null;
-    });
-
-    it('should return null when read is called', () => {
-      expect(iterator.read()).to.be.null;
-    });
-  });
-
   describe('An ArrayIterator with an empty array without autoStart', () => {
```
RubenVerborgh, line 46:
```suggestion
  describe('An ArrayIterator with an empty array', () => {
```

RubenVerborgh (commented):
Looking good!

I think we just need to add `autoStart` to all implementations now. So basically any class deriving from `AsyncIterator` (except from AsyncIterator itself, it seems).

Or, instead of adding the `autoStart` option, we need to change the behavior of `SingletonIterator` to just not automatically start (which is now the default).

jeswr:
> I think we just need to add autoStart to all implementations now. So basically any class deriving from AsyncIterator (except from AsyncIterator itself, it seems).

In order to handle https://github.com/RubenVerborgh/AsyncIterator/issues/35#issuecomment-1040187696 both solutions essentially are determining whether or not to autoStart the `end` event when there is no data so I think we will need to add it to `AsyncIterator` as well if we also want to solve this issue in this PR.

Also FYI @rubensworks with the current status of the PR only 15 tests from 7 test suites fail in the whole of comunica. Most of these issues arise from `metadata` tests in which `on('metadata', ...)` is called instead of `on('data', ...)` - this could either be solved directly in Comunica - or in this PR by making the `waitForDataListener` more generic and able to trigger flow mode on other event types.

rubensworks:
> Also FYI @rubensworks with the current status of the PR only 15 tests from 7 test suites fail in the whole of comunica. Most of these issues arise from metadata tests in which on('metadata', ...) is called instead of on('data', ...) - this could either be solved directly in Comunica - or in this PR by making the waitForDataListener more generic and able to trigger flow mode on other event types.

I'm fine with either option, whatever is easiest.
The `on('metadata', ...)` calls in Comunica are usually pretty deep internally, so it's ok if we have to do some additional things in Comunica here. (as long as we properly document this in the code)

Issue #37, opened 2022-02-19
Add toArray method


rubensworks:
While conserving the stream should be encouraged, there are still many use cases where converting the stream to an array is needed.
This is often done using arrayify-stream now, but this introduced the overhead of installing an additional package.
For developers, it would simplify things a lot if there was an (async) `toArray` method available, which would fully consume the stream and produce an array.

@RubenVerborgh If this sounds good to you, I will look into implementing this for including in the next major release (for #36).

RubenVerborgh:
Great, absolutely! (Might take an optional object with limit parameter for infinite iterators?)

jeswr:
@rubensworks - agreed this would be very useful; so long as you are just using consuming the data using the `data` and `end` events you should be able to do this without creating conflicts with what I am doing to resolve #35 (don't branch of #36 as a lot of those changes are not included in what I'm doing, just go off the main branch instead).

P.S. sorry for the delay - I'll push what I have today; editing some of the tests have been a bit time consuming. Once I've done that (which includes a solution to #38 ) it may be worth having a custom `toArray` method for the ArrayIterator that just returns the internal array? This may be useful in Comunica if an iterable (for instance of reasoning rules) is being pipelined through a set of actors each of which call `toArray` and work with the array and then create a new iterator using `fromArray` which they pass to the next actor.

rubensworks:
> it may be worth having a custom toArray method for the ArrayIterator that just returns the internal array?

Not sure. I had in mind that this method would arrayify elements that have not been consumed yet. E.g., if the user has already called `read()` several times, then I would expect these elements to not be included in the resulting array.
But a sliced version of the array might be possible though (but I would leave that as a possible optimization if needed).

PR #40, opened 2022-02-21
Add toArray method on AsyncIterator

rubensworks:
As discussed in #37, this adds an async method for collecting items in an array.


asynciterator.ts
```diff
@@ -309,6 +309,35 @@ export class AsyncIterator<T> extends EventEmitter {
...
+
+        // Resolve earlier if a limit was set and we have reached that limit
+        if (options?.limit && items.length === options.limit) {
+          resolved = true;
+          resolve(items);
+        }
+      });
+      this.on('end', () => {
+        if (!resolved)
+          resolve(items);
```
```suggestion
        resolve(items);
```

(and no need to keep the boolean; this exact mechanism is already present in `Promise`)

asynciterator.ts
```diff
@@ -309,6 +309,35 @@ export class AsyncIterator<T> extends EventEmitter {
...
+      let resolved = false;
+
+      this.on('error', reject);
+      this.on('data', item => {
+        if (!resolved)
+          items.push(item);
+
+        // Resolve earlier if a limit was set and we have reached that limit
+        if (options?.limit && items.length === options.limit) {
+          resolved = true;
```
```suggestion
        items.push(item);

        // Resolve earlier if a limit was set and we have reached that limit
        if (options?.limit && items.length === options.limit) {
          // Remove listeners here instead (so making them named functions)
```

asynciterator.ts
```diff
@@ -309,6 +309,35 @@ export class AsyncIterator<T> extends EventEmitter {
     return '';
   }
 
+  /**
+    Consume all remaining items of the iterator into an array that will be returned asynchronously.
+    @param {object} [options] Settings for array creation
+    @param {integer} [options.limit] The maximum number of items to place in the array.
+   */
+  toArray(options?: { limit?: number }): Promise<T[]> {
```
RubenVerborgh, line 317:
```suggestion
  toArray(options?: { limit?: number }): Promise<T[]> {
    const items: T[] = [];
    if (!(options?.limit > 0))
      return Promise.resolve(items);
```

FYI:
```javascript
[
    -5,    -1,        -0,
     0,   NaN, undefined,
  null, false,     "foo",
   0.1,     1,         5,    // only this line should fail
].map(limit => !(limit > 0))
// [
//   true,  true,  true,
//   true,  true,  true,
//   true,  true,  true,
//   false, false, false,   // OMG
// ]
```

asynciterator.ts
```diff
@@ -309,6 +309,35 @@ export class AsyncIterator<T> extends EventEmitter {
...
+      const items: T[] = [];
+      let resolved = false;
+
+      this.on('error', reject);
+      this.on('data', item => {
+        if (!resolved)
+          items.push(item);
+
+        // Resolve earlier if a limit was set and we have reached that limit
+        if (options?.limit && items.length === options.limit) {
```
```suggestion
        if (items.length > options?.limit) {
```
wouldn't want to dream someone setting a float limit, but yeah 😅

```javascript
[
    -5,    -1,        -0,
     0,   NaN, undefined,
  null, false,     "foo",
   0.1,     1,         3,
     4,     4.5,       6, // only this line should succeed
].map(limit => limit > 3)
// [
//   false, false, false,
//   false, false, false,
//   false, false, false,
//   false, false, false,
//   true,  true,  true,
// ]
```

test/AsyncIterator-test.js
```diff
@@ -1081,4 +1081,154 @@ describe('AsyncIterator', () => {
...
+          done();
+        }).catch(done);
+      });
+
+      it('should return an array with five elements', () => {
+        expect(result).deep.to.equal([1, 2, 3, 4, 5]);
+      });
+    });
+
+    describe('called on an iterator with five items with limit 0', () => {
```
RubenVerborgh, line 1180: Shall we test the (positive integer) limit cases with infinite sequences? This ensures nothing blows up.

test/AsyncIterator-test.js
```diff
@@ -1081,4 +1081,154 @@ describe('AsyncIterator', () => {
...
+        iterator.readable = true;
+        iterator.read = () => i++ < 5 ? i : (iterator.close() || null);
+        iterator.toArray({ limit: 0 }).then(array => {
+          result = array;
+          done();
+        }).catch(done);
+      });
+
+      it('should return an array with five elements', () => {
+        expect(result).deep.to.equal([1, 2, 3, 4, 5]);
```
RubenVerborgh, line 1248: So that should be the empty array for me.

test/AsyncIterator-test.js
```diff
@@ -1081,4 +1081,154 @@ describe('AsyncIterator', () => {
...
+        iterator.toArray({ limit: 3 }).then(array => {
+          result = array;
+          done();
+        }).catch(done);
+      });
+
+      it('should return an array with three elements', () => {
+        expect(result).deep.to.equal([1, 2, 3]);
+      });
+    });
```
RubenVerborgh, line 1290: Also test that not too much was read from the underlying source iterator.

asynciterator.ts
rubensworks: I did it like this because the TS compiler complains about comparing with `undefined`.
Should I suppress it with `@ts-ignore`, or keep the truthy `options?.limit` check?

rubensworks: Removing `options?.limit` made it simpler to implement the zero-limit behaviour, so I went with the `@ts-ignore` for now. Can still change it if needed.

RubenVerborgh: Yeah; I think TypeScript is 100% right to warn here, because there are many edge cases. However, we have checked them all, and we have the right behavior.

asynciterator.ts
```diff
@@ -309,6 +309,35 @@ export class AsyncIterator<T> extends EventEmitter {
...
+      const dataListener = (item: T) => {
+        // Resolve earlier if a limit was set and we have reached that limit
+        // @ts-ignore
+        if (items.length >= options?.limit) {
+          this.removeListener('data', dataListener);
+          resolve(items);
+        }
+        else {
+          items.push(item);
+        }
```
RubenVerborgh: The conditional pushing is not going to work; we will be reading 1 too many elements time. So we will need the old logic and the zero check on top.

asynciterator.ts
```diff
@@ -309,6 +309,35 @@ export class AsyncIterator<T> extends EventEmitter {
...
+  toArray(options?: { limit?: number }): Promise<T[]> {
+    return new Promise<T[]>((resolve, reject) => {
+      // Collect items in an array
+      const items: T[] = [];
+
+      // Create a named listener so we can easily remove it later
+      const dataListener = (item: T) => {
+        // Resolve earlier if a limit was set and we have reached that limit
+        // @ts-ignore
+        if (items.length >= options?.limit) {
```
RubenVerborgh: The `>=` is dangerous, as the `=` part will result in nullish coalescence:

```
[
    -5,    -1,        -0,
     0,   NaN, undefined,
  null, false,     "foo",
   0.1,     1,         3,
     4,     4.5,       6, // only this line should fail
].map(limit => 3 >= limit)
// [
//   true,  true,  true,
//   true,  false, false,
//   true,  true,  false,
//   true,  true,  true,
//   false, false, false
// ]
```

So we really need `!(index < limit)`.

asynciterator.ts
rubensworks: I think we have a different understanding of what an undefined limit should do here.
I think it should just consume _all_ items in the stream, while this code would always produce an empty array for an undefined limit.
I think dumping all items in an array will probably be the most common case, so I guess we should enable that?

asynciterator.ts
rubensworks: I think this is the behaviour we want though, see my comment above: https://github.com/RubenVerborgh/AsyncIterator/pull/40#discussion_r811820332

I handled all the cases in the unit tests (I think), and these should produce the expected results.

asynciterator.ts
RubenVerborgh: Iterators can be infinite or at least huge.

But indeed, an `undefined`/`null`/`+Infinity` limit should indeed not limit; my bad.

asynciterator.ts
RubenVerborgh: Right yes, once we are here, we know that there is no zero limit in place.

asynciterator.ts
```diff
@@ -309,6 +309,39 @@ export class AsyncIterator<T> extends EventEmitter {
...
+    @param {integer} [options.limit] The maximum number of items to place in the array.
+   */
+  toArray(options?: { limit?: number }): Promise<T[]> {
+    return new Promise<T[]>((resolve, reject) => {
+      // Collect items in an array
+      const items: T[] = [];
+
+      // Early return if the limit is invalid
+      // @ts-ignore
+      if (options?.limit === 0)
```
RubenVerborgh: Or negative (specifically `-0`). `limit` might be the result of a calculation so devs might not always see.

RubenVerborgh (commented):
Thanks, looking good. Just the case where limits are negative needs to be covered.

Suggestion: define a `const limit = typeof options?.limit === 'number' ? options?.limit : Infinity` and reuse.

rubensworks:
All comments should be resolved now @RubenVerborgh.

(you may want to use the squash-based merge, as my last commits probably shouldn't exist separately)

RubenVerborgh:
Thanks so much!

PR #41, opened 2022-02-23
DRAFT: autostart removal

jeswr:
This is still WIP and not ready for review - in particular most `CloneIterator` tests are still failing.

Just wanted to make it clear that I haven't stopped work on this, it was just a bit (lot) more fiddly to get working than I had originally anticipated.

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -81,6 +81,46 @@ export class AsyncIterator<T> extends EventEmitter {
   private _readable = false;
   protected _properties?: { [name: string]: any };
   protected _propertyCallbacks?: { [name: string]: [(value: any) => void] };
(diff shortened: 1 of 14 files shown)


asynciterator.ts
```diff
@@ -81,6 +81,46 @@ export class AsyncIterator<T> extends EventEmitter {
   private _readable = false;
   protected _properties?: { [name: string]: any };
   protected _propertyCallbacks?: { [name: string]: [(value: any) => void] };
+  private _readableFlowing: boolean | null = null;
```
RubenVerborgh, line 84: I wonder if it is a necessity to maintain this flag, but let's keep it for now and see if we can remove/simplify it after the unit tests pass.

PR #43, opened 2022-03-13
BREAKING: Call _end via read method

### LINKED ISSUE #25: Consider removing autoStart, or at least disabling it by default

jeswr:
resolves #25 #35


jeswr:
@rubensworks - in your implementation of `toArray` you removed the listeners rather than destroying the iterator when the limit was reached - is there a particular reason for this?

You will see that I have changed this behavior in https://github.com/RubenVerborgh/AsyncIterator/pull/43/commits/ed8d4918ae676e2fa09f12e8e147885ce2ce6f68 - this is because as discussed [here](https://github.com/RubenVerborgh/AsyncIterator/issues/35#issuecomment-1044795133) removing a 'data' listener should not pause the stream, so in the case of your tests with infinite iterators the test suite was entering an infinite loop. An alternative is to call `.pause` inside `toArray`; but I would hazard against this as it would give unexpected behavior if other data listeners are attached.

RubenVerborgh:
@jeswr Thanks for your work.

There are quite some unrelated changes in this PR, which makes it hard for me to assess. Could we do this one step by step, possibly with different PRs for things like linting etc.? Changes like [this one](https://github.com/RubenVerborgh/AsyncIterator/pull/43/files#diff-8c1065e179bf8c72bbaa6ae735f3be78aac1e609456c0086be7d2b3c1dae66ecR630) are a bit distracting; I'd want to focus on one thing at a time.

rubensworks:
> @rubensworks - in your implementation of toArray you removed the listeners rather than destroying the iterator when the limit was reached - is there a particular reason for this?

The reasoning here was that the iterator could still be used after calling `toArray()`. For example for splicing of multiple arrays from the iterator, or perhaps even resuming the stream after first consuming an array.
But given the discussion in https://github.com/comunica/comunica/issues/950, it might make sense to change this though.

RubenVerborgh:
In general, I like to leave options open. So the idea is that, after `toArray` is finished. the derived iterator has stopped existing and the source iterator is free again. This default enables cases such as the ones @rubensworks describes.

But `destroySource: true` might be appropriate.

PR #45, opened 2022-03-14
breaking: end on read (part 1)

### LINKED ISSUE #25: Consider removing autoStart, or at least disabling it by default

### LINKED ISSUE #35: [Possible bug] Empty iterator does not work after `await`

jeswr:
resolves https://github.com/RubenVerborgh/AsyncIterator/issues/25
resolves https://github.com/RubenVerborgh/AsyncIterator/issues/35

Part of https://github.com/RubenVerborgh/AsyncIterator/pull/43 without any lint or limit changes

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -98,19 +98,12 @@ export class AsyncIterator<T> extends EventEmitter {
     @returns {boolean} Whether the state was changed
     @emits module:asynciterator.AsyncIterator.end
   */
-  protected _changeState(newState: number, eventAsync = false) {
+  protected _changeState(newState: number) {
     // Validate the state change
     const valid = newState > this._state && this._state < ENDED;
-    if (valid) {
+    if (valid)
       this._state = newState;
-      // Emit the `end` event when changing to ENDED
-      if (newState === ENDED) {
-        if (!eventAsync)
-          this.emit('end');
-        else
-          taskScheduler(() => this.emit('end'));
-      }
-    }
+
     return valid;
   }
 
@@ -128,6 +121,8 @@ export class AsyncIterator<T> extends EventEmitter {
     @returns {object?} The next item, or `null` if none is available
   */
   read(): T | null {
+    if (this._state === CLOSED)
+      this._end();
     return null;
   }
 
@@ -170,8 +165,12 @@ export class AsyncIterator<T> extends EventEmitter {
     @emits module:asynciterator.AsyncIterator.end
   */
   close() {
-    if (this._changeState(CLOSED))
-      this._endAsync();
+    if (this._changeState(CLOSED)) {
+      if (this.readable)
+        emitData.call(this);
+      else
+        this.readable = true;
+    }
   }
 
   /**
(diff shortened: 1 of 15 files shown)


asynciterator.ts
```diff
@@ -1405,15 +1384,29 @@ export class MultiTransformIterator<S, D = S> extends TransformIterator<S, D> {
     // Remove transformers that have ended
     const transformerQueue = this._transformerQueue, optional = this._optional;
     let head, item;
-    while ((head = transformerQueue[0]) && head.transformer.done) {
+    while ((head = transformerQueue[0]) !== undefined) {
```
jeswr, line 1387: I feel the changes here could do with a little explanation:

In short - since `end` is now triggered by calling `.read`, we need to make sure that we now always have triggered at least one `.read` call before checking if the iterator is `.done`.

At the same time I added the ability to add data from multiple consecutive Transformers in the queue so long as they have data ready; this can reduce excessive calls to `_read` if the transformers in the queue only produce a small number of elements each.

jeswr:
@RubenVerborgh this is ready for review - apparently I don't have permissions to set you as a reviewer over here ...

RubenVerborgh:
Copying some test code by @jeswr from #75 

```
        // TODO: Re-enable this section check for once https://github.com/RubenVerborgh/AsyncIterator/pull/45
        // is merged in order to check that destruction occurs immediately rather than over several ticks
        // it('should destroy when closed before being read after map', async () => {
        //   iterator.map(x => x).close();
        //   // TODO: Re-enable this same-tick check once https://github.com/RubenVerborgh/AsyncIterator/pull/45
        //   // is merged
        //   // iterator.destroyed.should.be.true;

        //   // Wait a tick
        //   await Promise.resolve();
        //   iterator.destroyed.should.be.true;
        //   console.log('b', iterator._state, iterator.open, iterator.closed, iterator.ended, iterator.destroyed)
        // });

        // it('should destroy when closed before being read after map then filter', async () => {
        //   const iter = iterator.map(x => x);
        //   iter.filter(x => true).close();

        //   // TODO: Re-enable this same-tick check once https://github.com/RubenVerborgh/AsyncIterator/pull/45
        //   // is merged
        //   // iterator.destroyed.should.be.true;
        //   // it.destroyed.should.be.true;

        //   // Wait a tick
        //   await Promise.resolve();
        //   iterator.destroyed.should.be.true;
        //   iter.destroyed.should.be.true;
        // });
```

jeswr:
Note to self: whenever I revisit this PR, I should add some more tests, and review the code, to make sure that `error` events are never emitted before the iterator is started.

Issue #44, opened 2022-03-13
optimization: Reduce overhead of SimpleTransformIterator


jeswr:
For basic methods like `.map` and `.filter`, on `AsyncIterator`, using the `SimpleTransformIterator` introduces uneccessary overhead via the `readAndTransformSimple` operator. To improve performance in applications like Comunica it may be worth creating custom iterators with simpler read methods; for instance a synchronous `MapIterator` could skip buffering entirely and just be implemented as something like

```ts
export class MapIterator<T, S> extends AsyncIterator<S> {
    ...
    read() {
       if ((item = source.read()) !== null)
          return this.map(item);
    }
}
```

@jacoscaz I suspect this is the cause of the potential slowdown you were encountering when doing a large amount of chaining.

jacoscaz:
Indeed, chained buffering steps might explain the slowdowns I’ve seen. Normally I would expect performance to slow down linearly with the number of chained transforms but often I see slowdowns that seem somewhat exponential. I had been under the impression that methods such as `map()` and `filter()` behaved as you described but, given this issue, I was either plain wrong or possibly remembering a previous version. This would be a great addition to `asynciterator` itself but we could do it separately if better for backward compat.

However, I do not have hard numbers at hand. @jeswr have you done some profiling on this?

rubensworks:
@jeswr Do you think think the slowdown is caused by the complexity of the code in `readAndTransformSimple`, or the fact that it fills an internal buffer? If the latter, just disabling the buffer on intermediary iterators might also improve overall performance.

jeswr:
@rubensworks I'd suspect you're correct in that the intermediary buffers are the main overhead. 

I was thinking that one way around this is to have a 'chained transform' which applies multiple map/filter/transform operations in one go, so long as each part of that chain is synchronous - is this equivalent to what you mean by disabling internal buffers?

jeswr:
@jacoscaz - no I have not done profiling, it was just an observation I made when digging through the code when working on a PR I made earlier

jacoscaz:
> I was thinking that one way around this is to have a 'chained transform' which applies multiple map/filter/transform operations in one go

Would this not require knowing what those transformations look like ahead of time? Unless, perhaps, one were to have all intermediate actors return a transform function and have the final actor concatenate these.

[...]

Issue #38, opened 2022-02-20
Optimize ArrayIterator


jeswr:
It currently uses `.shift` which is a O(n) operation. So it costs O(n^2) to iterator over.

jeswr:
The BufferedIterator will also suffer similar performance issues when the maxBufferSize, as does the `MultiTransformer` with a large transformerQueue.

RubenVerborgh:
All major JavaScript engines implement behavior that is closer to O(1), see:
- https://lannonbr.com/blog/2020-01-27-shift-optimizations
- https://bugzilla.mozilla.org/show_bug.cgi?id=1348772
- https://github.com/frida/v8/blob/main/src/builtins/array-shift.tq

Trying to be faster ourselves will most likely have the opposite effect.

RubenVerborgh:
Given the discussion in #44, we might want to see if alternatively implementations (notably a linked list) would speed this up.

jacoscaz:
@RubenVerborgh interesting you should say that... I thought exactly the same, and came up with https://github.com/jacoscaz/AsyncIterator/blob/42311b68988a971fc072c896100d7082b4162470/asynciterator.ts#L744-L783 . I'm still testing it. Interested in a PR?

RubenVerborgh:
That's quite perfect and exactly what I had in mind!
Curious about perf; PR welcome!

(Nit: perhaps we should have that code in a separate file and include it.)

PR #46, opened 2022-03-24
Backs BufferedIterator with a linked-list instead of an array

jacoscaz:
While researching and testing potential improvements for #44 and #38 I stumbled into a ~10% performance improvement from changing `BufferedIterator` so that it is backed by a linked-list instead of an array. 

This should not be the case, given https://github.com/RubenVerborgh/AsyncIterator/issues/38#issuecomment-1047549674:

> All major JavaScript engines implement behavior that is closer to O(1), see:
> 
> https://lannonbr.com/blog/2020-01-27-shift-optimizations
> https://bugzilla.mozilla.org/show_bug.cgi?id=1348772
> https://github.com/frida/v8/blob/main/src/builtins/array-shift.tq

However, it does seem to be the case in Node.js 14.x, 16.x and 17.x based on experimental data. @jeswr has raised the issue on the V8 repo https://bugs.chromium.org/p/v8/issues/detail?id=12730 .


RubenVerborgh:
Great work, thanks!

I'm surprised it would only be 10% though, if the non-const behavior of Array shifting is indeed what is holding us back.
What measurements do you have?

jacoscaz:
I'm timing the following while consuming a pre-generated array of `200000` items:

```typescript
const elapsed = await time((arr) => {
    let iterator: AsyncIterator<number> = new RangeIterator(arr);
    iterator = new TransformIterator(iterator);
    iterator = new TransformIterator(iterator);
    iterator = new TransformIterator(iterator);
    iterator = new TransformIterator(iterator);
    iterator = new TransformIterator(iterator);
    iterator = new TransformIterator(iterator);
    iterator = new TransformIterator(iterator);
    iterator = new TransformIterator(iterator);
    iterator = new TransformIterator(iterator);
    iterator = new TransformIterator(iterator);
    return iterator;
  });
```

I seem to be consistently getting ~580ms for the NPM version (array) and ~530ms for this version (linked-list).

> I'm surprised it would only be 10% though, if the non-const behavior of Array shifting is indeed what is holding us back.
> What measurements do you have?

I am surprised by this, too. I had expected a significantly larger difference. Likely this is just one factor in a number of factors that, when combined, result in more appreciable gains. Happy with leaving this open for a while as we do more testing.

EDIT: I'm testing on a 2020 13'' MacBook Pro (Apple Silicon, M1, 16 GB RAM).

RubenVerborgh:
Thanks! Yes, more testing material would be useful. I think this particular case might just get highly optimized by V8, can we try it with some of the cases of the other issue?

Minor: I wonder if instead of `next?: LinkedNode<V>`, we might want to keep it monomorphic by having `next: LinkedNode<V>` where `tail` points to itself. (In any case I'd recommend `null` over `undefined`.)

jacoscaz:
> In any case I'd recommend null over undefined.

Done!

>  I wonder if instead of next?: LinkedNode<V>, we might want to keep it monomorphic by having next: LinkedNode<V> where tail points to itself.

How would you do this? Can you even initialize a JS object with a property that points to itself?

RubenVerborgh:
@jacoscaz I have no idea where the 10% came from, but I currently have

```javascript
console.time('BufferedIterator');
const it = new BufferedIterator();
for (let i = 0; i < 500_000; i++)
  it._push(i);
it.close();

it.on('data', () => null)
  .on('end', () => console.timeEnd('BufferedIterator'));
```

running at 33.9s with the array implementation, and 0.070s with your linked list, which is a 500x speed-up 😄

jacoscaz:
>  which is a 500x speed-up

Ha! Nice! I must have messed up something in my testing. This is coherent with the findings in #47 . Thanks for having a look and merging so quickly!

RubenVerborgh:
Thanks! This PR is now part of v3.4.0.

PR #48, opened 2022-03-24
Simpler and faster filtering, mapping, skipping and limiting

jacoscaz:
While researching #44 , @jeswr discovered that the performance of `.map()` and `.filter()` could be greatly improved by adding simple dedicated iterators for mapping and filtering, respectively. The impact of using these simplified iterators grows with the length of the chain of iterators.

As measured on a 13'' 2020 MacBook Pro (Apple Silicon, M1, 16 GB RAM), this PR makes the following test run in ~30ms. Before this PR the same test would terminate after ~300ms - a 10x difference.

```typescript
let i = 0;

class RangeIterator<O> extends AsyncIterator<O> {
  constructor(source: O[]) {
    super();
    let i = 0;
    let l = source.length;
    this.read = (): O | null => {
      if (i >= l) {
        this.close();
        return null;
      }
      return source[i++];
    };
    Promise.resolve().then(() => {
      this.emit('readable');
    });
  }
}

const iterator = new RangeIterator(arr)
  .map((item: number) => item)
  .map((item: number) => item)
  .map((item: number) => item)
  .map((item: number) => item)
  .map((item: number) => item)
  .map((item: number) => item)
  .map((item: number) => item)
  .map((item: number) => item)
  .map((item: number) => item)
  .map((item: number) => item);

const now = Date.now();
iterator.on('data', () => {}).on('end', () => {
});
```

Note that the `RangeIterator` class is just a stopgap for #47 .

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -456,7 +456,8 @@ export class AsyncIterator<T> extends EventEmitter {
     @returns {module:asynciterator.AsyncIterator} A new iterator that maps the items from this iterator
   */
   map<D>(map: (item: T) => D, self?: any): AsyncIterator<D> {
-    return this.transform({ map: self ? map.bind(self) : map });
+    // return this.transform({ map: self ? map.bind(self) : map });
+    return new MappingIterator(this, self ? map.bind(self) : map);
   }
 
   /**
@@ -469,7 +470,8 @@ export class AsyncIterator<T> extends EventEmitter {
   filter<K extends T>(filter: (item: T) => item is K, self?: any): AsyncIterator<K>;
   filter(filter: (item: T) => boolean, self?: any): AsyncIterator<T>;
   filter(filter: (item: T) => boolean, self?: any): AsyncIterator<T> {
-    return this.transform({ filter: self ? filter.bind(self) : filter });
+    return new FilteringIterator(this, self ? filter.bind(self) : filter);
+    // return this.transform({ filter: self ? filter.bind(self) : filter });
   }
 
   /**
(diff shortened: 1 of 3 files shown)


jacoscaz:
@RubenVerborgh if you like where this is going I can spend some time refining those new iterators (cleaning up event handlers, implementing `read()` as an actual class method, ...).

asynciterator.ts
```diff
@@ -1214,6 +1218,92 @@ function destinationFillBuffer<S>(this: InternalSource<S>) {
...
+  }
+}
+
+export class SkippingIterator<T> extends AsyncIterator<T> {
+  constructor(source: AsyncIterator<T>, skip: number) {
+    super();
+    let item: T | null;
+    let skipped = 0;
+    this.read = (): T | null => {
+      while ((item = source.read()) !== null) {
```
jeswr: Rather than having all of this logic once skipping is finished why not modify the `read` method once you are done skipping?

```ts
   this.read = (): T | null => {
      while ((item = source.read()) !== null) {
        skipped += 1;
        if (skipped === skip) {
           this.read = () => source.read();
           return item;
        }
      }
      return null;
    };
```

jeswr: Also doesn't holding the item in the global scope mean that you are unnecessarily persisting it in memory after it has been read?

jacoscaz: > why not modify the read method once you are done skipping

Agreed! I do something similar here https://github.com/belayeng/quadstore/blob/2a538b729df63351dc587bc15b83cc7283cba9d2/lib/get/sortingIterator.ts#L45 . However, setting methods in the constructor and/or messing with prototypal methods in the constructor is often frowned upon in TS codebases and I don't know where @RubenVerborgh stands on this.

> Also doesn't holding the item in the global scope mean that you are unnecessarily persisting it in memory after it has been read?

Yes. This is a habit of mine that dates back to when doing so actually made for a performance difference on some platforms. By now I think every platform must have optimized that away and this is probably just a bad habit to have.

RubenVerborgh: Monkey patching methods is indeed not a good idea in general.

RubenVerborgh:
> if you like where this is going I can spend some time refining those new iterators (cleaning up event handlers, implementing read() as an actual class method, ...).

@jacoscaz Yes please. However, is this still relevant after https://github.com/RubenVerborgh/AsyncIterator/pull/50?

jacoscaz:
#50 is still about `TransformIterator`, which extends `BufferedIterator`. The gains in #46 will make the impact of this PR less significant for sure but there should still be a difference between buffering and no buffering. I'll do some comparison between the state of things now that #46 has been merged and this PR (and also #49 ) and report back.

jacoscaz:
With main as it is right now, after merging #46 , going through the following chain takes ~210 ms with an array of 200k items  (MacBook Pro, Apple Silicon, M1, 16 GB RAM). With this PR that goes down to ~30 ms, so roughly 7x faster.

```typescript
new ArrayIterator(arr)
      .map((item: number) => item)
      .map((item: number) => item)
      .map((item: number) => item)
      .map((item: number) => item)
      .map((item: number) => item)
      .map((item: number) => item)
      .map((item: number) => item)
      .map((item: number) => item)
      .map((item: number) => item)
      .map((item: number) => item)
```

jacoscaz:
@RubenVerborgh this should be in a better place now. Those new iterators likely deserve dedicated tests but before that, should anything more be done In `_destroy()` when it comes to event handling?

jacoscaz:
This PR now includes @jeswr work from #53 , focused on combining subsequent `map()` and `filter()` operations into one single iterator.

jacoscaz:
Superseded by #57

PR #49, opened 2022-03-25
Simpler and faster wrapping

jacoscaz:
This PR aims to extend the gains found in #48 to the `wrap` function, skipping any kind of buffering when possible (i.e. when no transform options are provided). @jeswr suggested this in  #44 .

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -2121,58 +2121,55 @@ class HistoryReader<T> {
   }
 }
 
-export interface WrappingIteratorOptions {
-  letIteratorThrough?: boolean;
-}
-
-export type PromiseLike<T> = Pick<Promise<T>, 'then' | 'catch'>;
+export type IteratorLike<T> = EventEmitter & {
+  on: (event: string | symbol, listener: (...args: any[]) => void) => AsyncIterator<T>; read: () => T | null };
 
 /* eslint-disable arrow-body-style */
-export const isPromiseLike = <T>(item: { [key: string]: any }): item is PromiseLike<T> => {
-  return isFunction(item.then) && isFunction(item.catch);
+export const isIteratorLike = <T>(item: { [key: string]: any }): item is IteratorLike<T> => {
+  return isFunction(item.on) && isFunction(item.read);
 };
 
-export type IteratorLike<T> = EventEmitter & { on: () => any; read: () => T | null };
+export const isIterator = <T>(item: { [key: string]: any }): item is Iterator<T> => {
+  return isFunction(item.next);
+};
 
-export const isIteratorLike = <T>(item: EventEmitter & { [key: string]: any }): item is IteratorLike<T> => {
-  return isFunction(item.on) && isFunction(item.read);
+export const isIterable = <T>(item: { [key: string]: any }): item is Iterable<T> => {
+  return Symbol.iterator in item;
 };
 
 export class WrappingIterator<T> extends AsyncIterator<T> {
   protected _source?: IteratorLike<T>;
 
-  constructor(sourceOrPromise: EventEmitter | PromiseLike<EventEmitter>, options: WrappingIteratorOptions = {}) {
+  constructor(sourceOrPromise: WrapSource<T> | Promise<WrapSource<T>>, options: WrapOptions = {}) {
     super();
-    if (options.letIteratorThrough === true && sourceOrPromise instanceof AsyncIterator)
-      return sourceOrPromise;
-    if (isPromiseLike(sourceOrPromise)) {
+    if (isPromise(sourceOrPromise)) {
       sourceOrPromise
         .then(source => {
-          WrappingIterator._wrapSource<T>(source, this);
+          WrappingIterator._wrapSource<T>(source, this, options);
         })
         .catch(err => {
           this.emit('error', err);
         });
-    } else {
-      WrappingIterator._wrapSource<T>(sourceOrPromise, this);
+    }
+    else {
+      WrappingIterator._wrapSource<T>(sourceOrPromise, this, options);
     }
   }
 
-  protected static _wrapSource<T>(source: EventEmitter, wrapped: WrappingIterator<T>) {
-    if (!isIteratorLike<T>(source)) {
-      taskScheduler(() => {
-        wrapped.emit('error', new Error(`Invalid source: ${source}`));
-      });
-      return;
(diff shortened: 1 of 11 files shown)


asynciterator.ts
```diff
@@ -1811,17 +1811,52 @@ class HistoryReader<T> {
   }
 }
 
+export class WrappingIterator<T> extends AsyncIterator<T> {
+  protected _source?: InternalSource<T>;
+
+  constructor(sourceOrPromise: AsyncIterator<T> | Promise<AsyncIterator<T>> | EventEmitter | Promise<EventEmitter>) {
+    super();
+    if (sourceOrPromise instanceof AsyncIterator)
+      return sourceOrPromise;
```
jeswr: As I believe you raised earlier; one of the notable things with `wrap`ing is that it will drop metadata. This modifies that behavior which may have unintended implications downstream.

jacoscaz: Just to make sure I understand, is the implication here that downstream objects might receive unexpected and potentially conflicting metadata?

Good catch! This would indeed be breaking behavior. Do you think it’d be worth having an option to allow this behavior, maybe set to false by default, or should we just wrap everything anyway?

jeswr: > Just to make sure I understand, is the implication here that downstream objects might receive unexpected and potentially conflicting metadata?

Yes

> Do you think it’d be worth having an option to allow this behaviour, maybe set to false by default, or should we just wrap everything anyway?

Might be a nice feature to have; but definitely false by default.

jacoscaz: @jeswr done!

RubenVerborgh:
Do we have any performance evidence here?

jacoscaz:
With main as it is right now, after merging https://github.com/RubenVerborgh/AsyncIterator/pull/46 , going through an array of 200k items with `wrap(new ArrayIterator(arr))` takes ~103 ms (MacBook Pro, Apple Silicon, M1, 16 GB RAM). With this PR that goes down to ~10 ms, so roughly 10x faster.

jacoscaz:
@jeswr this PR now includes your work from #55 , as discussed on Gitter.

@jeswr @RubenVerborgh I've tried reworking the `wrap()` function to return the most appropriate iterator for all sorts of sources while maintaining backward compat. with the current implementation. Seems to work well but looking forward to your feedbacks. This PR now builds on #48 , diff here https://github.com/jacoscaz/AsyncIterator/compare/faster-filtering-mapping...jacoscaz:faster-wrapping .

jacoscaz:
I realize that the resulting PR might get pretty big but perhaps we could merge this one into #48 so as to have all of the recent performance work in one branch. That would simplify testing in other projects.

jeswr:
I think the _wrap function still needs an option where you can select to run the isIterable tests first within the wrap function.

I'm also concerned that this is a little heavyweight if you do lots of wrapping, for instance within the map or transform of another iterator. I think it's worth having custom wrap functions for each of those source types that the user can choose to use if they know the nature of what they are wrapping in advance.

Great work on this btw @jacoscaz !

jacoscaz:
@jeswr I should have addressed both points in my latest commit. The names are a bit confusing, though, particularly the `fromIterator` function (for ES2015 `Iterator`) and `fromIteratorLike` function (for `AsyncIterator`, `stream.Readable`, ...). Suggestions?

jeswr:
@jacoscaz , all of the `iteratorLike` stuff could potentially be `AsyncIteratorLike`? Because the `fromIterator` is strictly for ES2015 synchronous iterables?

jacoscaz:
All right, done! Let's see what @RubenVerborgh says; if things look good I'll add tests for both this and #48 and we should be all set.

asynciterator.ts
```diff
@@ -1251,6 +1259,244 @@ function destinationFillBuffer<S>(this: InternalSource<S>) {
...
+  filter(filter: (item: D) => boolean, self?: any): AsyncIterator<D> {
+    return new MultiMapFilterTransformIterator(this._source, {
+      filter: true,
+      function: self ? filter.bind(self) : filter,
+      next: {
+        filter: false,
+        function: this._map,
+      },
+    });
+  }
```
jeswr, line 1374: Per [this comment](https://github.com/RubenVerborgh/AsyncIterator/pull/53#issuecomment-1079812777) it still may be worth removing these lines. @jacoscaz it might be worth doing some perf. tests on your machine to see if your results align with mine.

jeswr: I think the cause of the marginal slowdown observed there was the fact that we are essentially adding an extra function for every map operation in order to compose them `(item: any) => t(_func(item));`

jacoscaz: Curiously enough, I am getting 10% - 15% _faster_ iteration by keeping those lines in. Removing them seems to slow things down. On some level this makes sense as V8 should optimize execution of the "compiled" transformation more effectively. Another differentiating factor might be the number of item vs. the number of iterators. These lines likely produce higher setup latencies, so they might have a more significant impact on scenarios with low numbers of items and high numbers of iterators. I would keep these in and postpone further tuning to a later stage as we're currently lacking dedicated scaffolding for this sort of testing.

jacoscaz:
Happy to report that not only all tests are passing but with this branch we're not that far from having all tests in Comunica pass, too! There's only 8 tests that are breaking. Issues so far:

- There's one test suite that passes a set of initial quads as an empty object, which breaks wrapping https://github.com/comunica/comunica/blob/448bfbadd805dcad95841789cd12b0d33a672585/packages/actor-rdf-resolve-hypermedia-qpf/test/ActorRdfResolveHypermediaQpf-test.ts#L103 . Fixing this one is easy, just change the test to make that an empty array.
- There's one actor that uses the protected method `BufferedIterator#_fillBuffer()`, which is not available anymore as after these PRs there's no guarantee for `wrap()` to return an instance of `BufferedIterator` (and we shouldn't use protected methods anyway) https://github.com/comunica/comunica/blob/448bfbadd805dcad95841789cd12b0d33a672585/packages/actor-query-operation-sparql-endpoint/lib/ActorQueryOperationSparqlEndpoint.ts#L168 . Fixing this one might not be as easy but I don't have a clear idea of what's going on.
- This test https://github.com/comunica/comunica/blob/448bfbadd805dcad95841789cd12b0d33a672585/packages/actor-query-operation-sparql-endpoint/test/ActorQueryOperationSparqlEndpoint-test.ts#L410 is breaking due to a timeout while expecting an `error` event. This might be due to something missing in how we relay errors within these PRs.

jacoscaz:
Superseded by #57

PR #50, opened 2022-03-26
Feat transform iterator

jeswr:
Per discussion in #44 -  a fast transform iterator that uses mutations. Happy to take suggestions for renaming.

> Could be a separate file though (and maybe we could also split the code into more separate files in general).

@RubenVerborgh  This requires modifying some of the package.json and tsconfig.json settings. So it may be better to do this in a separate PR.

This code could probably do with some more integration tests in the future - but not sure if it is worth bothering until the changes in #45 get integrated.

Pinging @jacoscaz


jeswr:
I *tried* to add some features for throwing extra errors/warnings in production mode in https://github.com/RubenVerborgh/AsyncIterator/pull/50/commits/3a57d17b832731f1e6781a00a9e030b243276473 - but I couldn't quite figure out how to configure the configs and test-suite properly to handle this.

RubenVerborgh:
@jeswr Thanks! Do we have any insights into the performance of this one?

asynciterator.ts
```diff
@@ -1923,3 +1923,116 @@ type SourceExpression<T> =
...
+      source.on('readable', () => {
+        this.emit('readable');
+      });
+      source.on('end', () => {
+        this.close();
+      });
+    }
+
+    read(): T | null {
+      const func = build(this.transforms);
```
RubenVerborgh, line 1955: We can cache this instead.

asynciterator.ts
```diff
@@ -1923,3 +1923,116 @@ type SourceExpression<T> =
...
+    map<D>(map: (item: T) => D): FastTransformIterator<D> {
+      this.transforms.push({ filter: false, function: map });
+      return this as unknown as FastTransformIterator<D>;
+    }
+
+    /**
+      Transforms items according to a synchronous generator (hence no need for buffering)
+      @param {Function} transform The function to transform items with
+    */
+    syncTransform<D>(transform: (item: T) => Generator<D>): FastTransformIterator<D> {
```
RubenVerborgh, line 1994: What are the use cases for this transformation? Is this just a multimap?

RubenVerborgh (commented):
So I like the idea, but I think mutation unnecessarily complicates things.

Suggestion: can we instead override `map` and `filter` methods on the original `TransformIterator`, with a check that _if_ the `source` already is `instanceof TransformIterator` that we create this more optimized iterator instead?

For instance, `source.map(x => x * 2).map(y => y + 1)`. Then the second `map` recognizes that the input is already a `TransformIterator`, so what it actually returns is a `TransformIterator` over `source` (!) with as map function the combination of the two mappings. So basically `source.map(combine(x => x * 2, y => y + 1))`.

asynciterator.ts
jeswr: Yep!

jeswr:
> @jeswr Thanks! Do we have any insights into the performance of this one?

On the experiments I did I saw up to a 2x improvement in some of the tests I did a few days ago when compared to the chained use of MapIterator and FilterIterator (I haven't done benchmarks for the transform). I'd imagine that the performance increase is somewhat proportional to the number of elements you are filtering out as well since this essentially eliminates any upstream `read`s that would have been made in these cases.

PR #53, opened 2022-03-27
Faster filtering mapping

jeswr:
Supercedes #48 
Supercedes #50 

Makes use of [this](https://github.com/RubenVerborgh/AsyncIterator/pull/50#pullrequestreview-922328793) suggestion by @RubenVerborgh to provide the ~2x speedup to chained maps and transforms; without having the problems associated with immutability (to give you a sense, the test suite is now running at ~600ms on my machine vs ~1000ms before).

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -456,7 +456,8 @@ export class AsyncIterator<T> extends EventEmitter {
     @returns {module:asynciterator.AsyncIterator} A new iterator that maps the items from this iterator
   */
   map<D>(map: (item: T) => D, self?: any): AsyncIterator<D> {
-    return this.transform({ map: self ? map.bind(self) : map });
+    // return this.transform({ map: self ? map.bind(self) : map });
+    return new MappingIterator(this, self ? map.bind(self) : map);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


jeswr:
The performance improvements greatly depend on the number of filters. In the example @jacoscaz had with only maps - there was no gain (if anything a slight degradation of 5-10%), but if there are filters involved the speedup can be much more significant. For instance running

```ts
range(0, 2_000_000)
.map((item: number) => item)
.filter((item: number) => item % 2 === 0)
.map((item: number) => item)
.filter((item: number) => item % 3 === 0)
.map((item: number) => item)
.filter((item: number) => item % 4 === 0)
.map((item: number) => item)
.filter((item: number) => item % 5 === 0)
.map((item: number) => item)
.filter((item: number) => item % 6 === 0)
```

Goes from ~80ms to ~20ms; so a 4x speedup

As a result - it may be worth only triggering this optimized iterator when we reach the first `FilterIterator` which we can do by simply removing the following lines https://github.com/jeswr/AsyncIterator/blob/1ec39fbb12497ad46ea5409ba683deb4cbb2079c/asynciterator.ts#L1331-L1351

jeswr:
I'm not sure I understand the cause of the failing test on node v10

jeswr:
As discussed on gitter @jacoscaz  will merge this into #48 - I will close this once that has been done.

jacoscaz:
Done!

PR #55, opened 2022-03-27
feat: wrapIterator

jeswr:
See https://github.com/comunica/comunica/issues/965 especially in light of https://github.com/rdfjs/N3.js/pull/275.

Once this and all of the other open PRs around fastIterators are closed, this should mean that Comunica will no longer store unecessary copies of matches in various arrays/buffers throughout the engine!

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -1852,6 +1852,22 @@ class HistoryReader<T> {
   }
 }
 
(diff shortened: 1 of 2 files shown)


jeswr:
This complements #49 - but does not override any of the performance gains it provides.

PR #58, opened 2022-04-04
feat: faster transforming

jeswr:
This acts as PR 1 defined in https://github.com/RubenVerborgh/AsyncIterator/issues/44#issuecomment-1086950726.

The updated and cleaned up implementation is approximately 2x faster than the original `pre-compiling` strategy of transformation.

This is ready for review @RubenVerborgh @jacoscaz

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -53,7 +53,7 @@
     no-implied-eval: error,
     no-invalid-this: off,
     no-iterator: error,
-    no-labels: error,
+    no-labels: off,
     no-lone-blocks: error,
     no-loop-func: error,
     no-magic-numbers: off,
--- a/.eslintrc
+++ b/.eslintrc
@@ -53,7 +53,7 @@
     no-implied-eval: error,
     no-invalid-this: off,
     no-iterator: error,
-    no-labels: off,
+    no-labels: error,
     no-lone-blocks: error,
     no-loop-func: error,
     no-magic-numbers: off,
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -1292,7 +1292,7 @@ export class MappingIterator<T, D = T> extends AsyncIterator<D> {
     upstream.on('error', this.onSourceError);
     upstream.on('readable', this.onSourceReadable);
     if (upstream.done)
-      this.onSourceEnd();
+      this.close();
   }
 
   get fn() {
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -1288,11 +1288,13 @@ export class MappingIterator<T, D = T> extends AsyncIterator<D> {
     // listeners to the original source
     super();
     this._destroySource = options.destroySource !== false;
-    upstream.on('end', this.onSourceEnd);
-    upstream.on('error', this.onSourceError);
-    upstream.on('readable', this.onSourceReadable);
     if (upstream.done)
       this.close();
+    else {
+      upstream.on('end', this.onSourceEnd);
+      upstream.on('error', this.onSourceError);
+      upstream.on('readable', this.onSourceReadable);
+    }
   }
 
   get fn() {
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -451,7 +451,8 @@ export class AsyncIterator<T> extends EventEmitter {
   /**
     Maps items from this iterator using the given function.
     After this operation, only read the returned iterator instead of the current one.
-    @param {Function} map A mapping function to call on this iterator's (remaining) items (null values are skipped)
+    @param {Function} map A mapping function to call on this iterator's (remaining) items.
+        A `null` value indicates that nothing should be returned for a particular item..
     @param {object?} self The `this` pointer for the mapping function
     @returns {module:asynciterator.AsyncIterator} A new iterator that maps the items from this iterator
   */
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -527,7 +527,7 @@ export class AsyncIterator<T> extends EventEmitter {
     @returns {module:asynciterator.AsyncIterator} A new iterator with at most the given number of items
   */
   take(limit: number): AsyncIterator<T> {
-    return new LimitingIterator(this, limit);
+    return new HeadIterator(this, limit);
   }
 
   /**
@@ -1376,7 +1376,7 @@ export class SyncTransformIterator<T, D = T> extends SynchronousTransformIterato
   }
 }
 
-export class LimitingIterator<T> extends SynchronousTransformIterator<T> {
+export class HeadIterator<T> extends SynchronousTransformIterator<T> {
   protected count: number = 0;
 
   constructor(source: AsyncIterator<T>, protected readonly limit: number) {
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -1255,9 +1255,11 @@ function destinationFillBuffer<S>(this: InternalSource<S>) {
 
 
 export abstract class SynchronousTransformIterator<S, D = S> extends AsyncIterator<D> {
-  protected constructor(protected _source: AsyncIterator<S>, private options: { destroySource?: boolean } = {}) {
+  private _destroySource: boolean;
+  protected constructor(protected _source: AsyncIterator<S>, protected options: { destroySource?: boolean } = {}) {
     /* eslint-disable no-use-before-define */
     super();
+    this._destroySource = options.destroySource !== false;
     const cleanup = () => {
       _source.removeListener('end', onSourceEnd);
       _source.removeListener('error', onSourceError);
@@ -1293,12 +1295,13 @@ export abstract class SynchronousTransformIterator<S, D = S> extends AsyncIterat
       onSourceReadable();
   }
 
-  protected _destroy(cause: Error | undefined, callback: (error?: Error) => void) {
-    super._destroy(cause, callback);
+  destroy(cause?: Error): void {
+    this._source.destroy(cause);
+    super.destroy(cause);
   }
 
   public close() {
-    if (this.options.destroySource)
+    if (this._destroySource)
       this._source.destroy();
     super.close();
   }
@@ -1312,10 +1315,10 @@ interface Transform {
 export class SyncTransformIterator<T, D = T> extends SynchronousTransformIterator<T, D> {
   private _funcs?: Function[];
 
-  constructor(private source: AsyncIterator<T>, private transforms: Transform, upstream: AsyncIterator<any> = source) {
+  constructor(private source: AsyncIterator<T>, private transforms: Transform, private upstream: AsyncIterator<any> = source, options?: { destroySource?: boolean }) {
     // Subscribe the iterator directly upstream rather than the original source to avoid over-subscribing
     // listeners to the original source
-    super(upstream);
+    super(upstream, options);
   }
 
   get funcs() {
@@ -1348,6 +1351,17 @@ export class SyncTransformIterator<T, D = T> extends SynchronousTransformIterato
 
   map<K>(map: (item: D) => K | null, self?: any): AsyncIterator<K> {
     return new SyncTransformIterator<T, K>(this.source, { fn: self ? map.bind(self) : map, next: this.transforms }, this);
+  }
+
+  destroy(cause?: Error): void {
+    this.upstream.destroy(cause);
+    super.destroy(cause);
+  }
+
+  public close() {
+    if (this.options.destroySource)
+      this.upstream.destroy();
+    super.close();
   }
 }
 
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -1312,15 +1312,15 @@ export abstract class SynchronousTransformIterator<S, D = S> extends AsyncIterat
   }
 }
 
-interface Transform {
+interface ComposedFunction {
   fn: Function,
-  next?: Transform
+  next?: ComposedFunction
 }
 
 export class SyncTransformIterator<T, D = T> extends SynchronousTransformIterator<T, D> {
   private _fn?: Function;
 
-  constructor(private source: AsyncIterator<T>, private transforms: Transform, private upstream: AsyncIterator<any> = source, options?: { destroySource?: boolean }) {
+  constructor(private source: AsyncIterator<T>, private transforms: ComposedFunction, private upstream: AsyncIterator<any> = source, options?: { destroySource?: boolean }) {
     // Subscribe the iterator directly upstream rather than the original source to avoid over-subscribing
     // listeners to the original source
     super(upstream, options);
@@ -1330,7 +1330,7 @@ export class SyncTransformIterator<T, D = T> extends SynchronousTransformIterato
     if (!this._fn) {
       const funcs: Function[] = [];
       // eslint-disable-next-line prefer-destructuring
-      let transforms: Transform | undefined = this.transforms;
+      let transforms: ComposedFunction | undefined = this.transforms;
       do
         funcs.push(transforms.fn);
       // eslint-disable-next-line no-cond-assign
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -1278,7 +1278,12 @@ export class MappingIterator<T, D = T> extends AsyncIterator<D> {
     this.source.readable = readable;
   }
 
-  constructor(protected source: AsyncIterator<T>, private transforms?: ComposedFunction, private upstream: AsyncIterator<any> = source, options: { destroySource?: boolean } = {}) {
+  constructor(
+    protected source: AsyncIterator<T>,
+    private transforms?: ComposedFunction,
+    private upstream: AsyncIterator<any> = source,
+    options: { destroySource?: boolean } = {}
+  ) {
     // Subscribe the iterator directly upstream rather than the original source to avoid over-subscribing
     // listeners to the original source
     super();
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -1273,9 +1273,11 @@ export class MappingIterator<T, D = T> extends AsyncIterator<D> {
     super();
     this._destroySource = options.destroySource !== false;
     const cleanup = () => {
+      /* eslint-disable no-use-before-define */
       upstream.removeListener('end', onSourceEnd);
       upstream.removeListener('error', onSourceError);
       upstream.removeListener('readable', onSourceReadable);
+      /* eslint-enable no-use-before-define */
     };
     const onSourceEnd = () => {
       cleanup();
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -25,6 +25,11 @@ export function setTaskScheduler(scheduler: TaskScheduler): void {
   taskScheduler = scheduler;
 }
 
+/** Binds a function to an object */
+function bind(fn: Function, self: any) {
+  return self ? fn.bind(self) : fn;
+}
+
 /**
   ID of the INIT state.
   An iterator is initializing if it is preparing main item generation.
@@ -470,8 +475,7 @@ export class AsyncIterator<T> extends EventEmitter {
   filter<K extends T>(filter: (item: T) => item is K, self?: any): AsyncIterator<K>;
   filter(filter: (item: T) => boolean, self?: any): AsyncIterator<T>;
   filter(filter: (item: T) => boolean, self?: any): AsyncIterator<T> {
-    if (self)
-      filter = filter.bind(self);
+    filter = bind(filter, self);
     return this.map(item => filter(item) ? item : null);
   }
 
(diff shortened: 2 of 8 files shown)


asynciterator.ts
```diff
@@ -1252,6 +1254,135 @@ function destinationFillBuffer<S>(this: InternalSource<S>) {
...
+export abstract class SynchronousTransformIterator<S, D = S> extends AsyncIterator<D> {
+  private _destroySource: boolean;
+  protected constructor(protected _source: AsyncIterator<S>, protected options: { destroySource?: boolean } = {}) {
+    /* eslint-disable no-use-before-define */
+    super();
+    this._destroySource = options.destroySource !== false;
+    const cleanup = () => {
+      _source.removeListener('end', onSourceEnd);
+      _source.removeListener('error', onSourceError);
+      _source.removeListener('readable', onSourceReadable);
```
jacoscaz: I would add a delete `this._source` here, possibly wrapped by `scheduleTask` so that leftover `.read()` operations do not break.

asynciterator.ts
```diff
@@ -1252,6 +1254,135 @@ function destinationFillBuffer<S>(this: InternalSource<S>) {
...
+    super.close();
+  }
+}
+
+interface Transform {
+  fn: Function,
+  next?: Transform
+}
+
+export class SyncTransformIterator<T, D = T> extends SynchronousTransformIterator<T, D> {
```
jacoscaz: `SyncTransformIterator` and `SynchronousTransformIterator` are too similar IMHO. Perhaps:

```
SynchronousTrasformIterator => BaseSynchronousIterator
SyncTransformIterator => SynchronousTransformIterator
```

asynciterator.ts
```diff
@@ -1252,6 +1254,135 @@ function destinationFillBuffer<S>(this: InternalSource<S>) {
...
+      for (let index = funcs.length - 1; index >= 0; index -= 1) {
+        if ((item = funcs[index](item)) === null)
+          continue outer;
+      }
+      return item;
+    }
+    return null;
+  }
+
+  map<K>(map: (item: D) => K | null, self?: any): AsyncIterator<K> {
```
jacoscaz, line 1335: The `null` here (I don't want this item to flow downstream) is semantically different from the `null` returned by `read()` (no items available). Can we replace this one with a `Symbol`?

jacoscaz (changes requested):
@jeswr this is an amazing take on the work done so far when it comes to synchronous transformations. Hats off to you!

jacoscaz:
Performance-wise, the following tests takes ~210 ms on my machine (MacBook Pro, Apple Silicon, M1, 16 GB) when using the `main` branch and ~23 ms when using this branch - 9.3x improvement.

```typescript
let i = 0;

const iterator = new ArrayIterator(arr)
  .map((item: number) => item)
  .map((item: number) => item)
  .map((item: number) => item)
  .map((item: number) => item)
  .map((item: number) => item)
  .map((item: number) => item)
  .map((item: number) => item)
  .map((item: number) => item)
  .map((item: number) => item)
  .map((item: number) => item);

const now = Date.now();
iterator.on('data', () => {}).on('end', () => {
});
```

.eslintrc
```diff
@@ -53,7 +53,7 @@
     no-implied-eval: error,
     no-invalid-this: off,
     no-iterator: error,
-    no-labels: error,
+    no-labels: off,
```
RubenVerborgh: https://homepages.cwi.nl/~storm/teaching/reader/Dijkstra68.pdf

asynciterator.ts
```diff
@@ -451,12 +451,12 @@ export class AsyncIterator<T> extends EventEmitter {
   /**
     Maps items from this iterator using the given function.
     After this operation, only read the returned iterator instead of the current one.
-    @param {Function} map A mapping function to call on this iterator's (remaining) items
+    @param {Function} map A mapping function to call on this iterator's (remaining) items (null values are skipped)
```
```suggestion
    @param {Function} map A mapping function to call on this iterator's (remaining) items (null values are skipped)
```
```suggestion
    @param {Function} map A mapping function to call on this iterator's (remaining) items.
        A `null` value indicates that nothing should be returned for a particular item..
```

```suggestion
    @param {Function} map A mapping function to call on this iterator's (remaining) items (null values are skipped)
```
```suggestion
    @param {Function} map A mapping function to call on this iterator's (remaining) items.
        A `null` value indicates that nothing should be returned for a particular item..
```

asynciterator.ts
```diff
@@ -451,12 +451,12 @@ export class AsyncIterator<T> extends EventEmitter {
...
     @param {object?} self The `this` pointer for the mapping function
     @returns {module:asynciterator.AsyncIterator} A new iterator that maps the items from this iterator
   */
-  map<D>(map: (item: T) => D, self?: any): AsyncIterator<D> {
-    return this.transform({ map: self ? map.bind(self) : map });
+  map<D>(map: (item: T) => D | null, self?: any): AsyncIterator<D> {
+    return new SyncTransformIterator<T, D>(this, { fn: self ? map.bind(self) : map });
```
RubenVerborgh: We probably want a helper function

```javascript
function bind (fn, self) {
  return self ? fn.bind(self) : fn;
}
```

asynciterator.ts
```diff
@@ -1252,6 +1254,135 @@ function destinationFillBuffer<S>(this: InternalSource<S>) {
 }
 
 
+export abstract class SynchronousTransformIterator<S, D = S> extends AsyncIterator<D> {
+  private _destroySource: boolean;
+  protected constructor(protected _source: AsyncIterator<S>, protected options: { destroySource?: boolean } = {}) {
+    /* eslint-disable no-use-before-define */
+    super();
+    this._destroySource = options.destroySource !== false;
+    const cleanup = () => {
```
RubenVerborgh: The constructor definitely needs some cleaning up; it's very hard to understand what's going on with all of the helper functions.

RubenVerborgh: The constructor definitely needs some cleaning up; it's very hard to understand what's going on with all of the helper functions.

asynciterator.ts
```diff
@@ -1252,6 +1254,135 @@ function destinationFillBuffer<S>(this: InternalSource<S>) {
 }
 
 
+export abstract class SynchronousTransformIterator<S, D = S> extends AsyncIterator<D> {
+  private _destroySource: boolean;
+  protected constructor(protected _source: AsyncIterator<S>, protected options: { destroySource?: boolean } = {}) {
```
```suggestion
  protected constructor(protected source: AsyncIterator<S>, options: { destroySource?: boolean } = {}) {
```

```suggestion
  public constructor(protected _source: AsyncIterator<S>, protected options: { destroySource?: boolean } = {}) {
```

asynciterator.ts
```diff
@@ -1252,6 +1254,135 @@ function destinationFillBuffer<S>(this: InternalSource<S>) {
 }
 
 
+export abstract class SynchronousTransformIterator<S, D = S> extends AsyncIterator<D> {
```
```suggestion
export abstract class MappedIterator<S, D = S> extends AsyncIterator<D> {
```

Let's just call it this.

The fact that the `Synchronous` pertains to `Transform` but not to `Iterator` is too hard to understand.

asynciterator.ts
```diff
@@ -1252,6 +1254,135 @@ function destinationFillBuffer<S>(this: InternalSource<S>) {
...
+    const onSourceEnd = () => {
+      cleanup();
+      this.close();
+    };
+    const onSourceError = (err: Error) => {
+      this.emit('error', err);
+    };
+    const onSourceReadable = () => {
+      if (this.readable) {
+        // TODO: I'm not completely sure as to why this is needed but without
```
RubenVerborgh: Needs fixing.

asynciterator.ts
```diff
@@ -1252,6 +1254,135 @@ function destinationFillBuffer<S>(this: InternalSource<S>) {
...
+  }
+
+  public close() {
+    if (this._destroySource)
+      this._source.destroy();
+    super.close();
+  }
+}
+
+interface Transform {
```
```suggestion
interface ComposedFunction {
```

asynciterator.ts
```diff
@@ -1252,6 +1254,135 @@ function destinationFillBuffer<S>(this: InternalSource<S>) {
...
+      let transforms: Transform | undefined = this.transforms;
+      do
+        this._funcs.push(transforms.fn);
+      // eslint-disable-next-line no-cond-assign
+      while (transforms = transforms.next);
+    }
+    return this._funcs;
+  }
+
+  read(): D | null {
```
RubenVerborgh, line 1325: This implementation is too funky.
I think we can do it without label jumping etc.

asynciterator.ts
```diff
@@ -1252,6 +1254,135 @@ function destinationFillBuffer<S>(this: InternalSource<S>) {
...
+    const { source, funcs } = this;
+    let item;
+    outer: while ((item = source.read()) !== null) {
+      // Do not use a for-of loop here, it slows down transformations
```
RubenVerborgh: What about a `reduceRight` loop, which is what this seems to be?
But I see we have an early stop condition indeed.

asynciterator.ts
```diff
@@ -1252,6 +1254,135 @@ function destinationFillBuffer<S>(this: InternalSource<S>) {
...
+export class SyncTransformIterator<T, D = T> extends SynchronousTransformIterator<T, D> {
+  private _funcs?: Function[];
+
+  constructor(private source: AsyncIterator<T>, private transforms: Transform, private upstream: AsyncIterator<any> = source, options?: { destroySource?: boolean }) {
+    // Subscribe the iterator directly upstream rather than the original source to avoid over-subscribing
+    // listeners to the original source
+    super(upstream, options);
+  }
+
+  get funcs() {
```
RubenVerborgh: Let's call them `mappings` or so.

But why the double choice of an array _and_ a linked list structure? We can just pass in the array in the constructor. I don't think we need `Transform`.

asynciterator.ts
RubenVerborgh: Let's just merge both classes.

asynciterator.ts
```diff
@@ -1252,6 +1254,135 @@ function destinationFillBuffer<S>(this: InternalSource<S>) {
...
+  }
+
+  public close() {
+    if (this.options.destroySource)
+      this.upstream.destroy();
+    super.close();
+  }
+}
+
+export class LimitingIterator<T> extends SynchronousTransformIterator<T> {
```
RubenVerborgh: Why not just directly from `AsyncIterator`, or `TransformIterator`?
All we need is the `source` mechanics.

test/AsyncIterator-test.js
```diff
@@ -1307,4 +1311,231 @@ describe('AsyncIterator', () => {
       });
     });
   });
+  describe('Testing chains fo maps and filters', () => {
```
RubenVerborgh:
```suggestion
  describe('A chain of maps and filters', () => {
```

RubenVerborgh: (throughout)

test/AsyncIterator-test.js
```diff
@@ -1307,4 +1311,231 @@ describe('AsyncIterator', () => {
...
+    for (const iteratorGen of [() => range(0, 2), () => fromArray([0, 1, 2]), () => wrap(range(0, 2))]) {
+      // eslint-disable-next-line no-loop-func
+      describe(`Testing with ${iteratorGen()}`, () => {
+        let iterator;
+        beforeEach(() => {
+          iterator = iteratorGen();
+        });
+        it('Should handle no transforms arrayified', async () => {
```
RubenVerborgh:
```suggestion

        it('should handle no transforms arrayified', async () => {
```

RubenVerborgh: (throughout: so newline and lowercase)

asynciterator.ts
```suggestion
export class HeadIterator<T> extends SynchronousTransformIterator<T> {
```
or
```suggestion
export class PartialIterator<T> extends SynchronousTransformIterator<T> {
```

RubenVerborgh (commented):
Thanks for your work!
The ideas are definitely good, but a significant amount of polishing is required.

This is a very tricky library; it is used in performance-critical code that is sensitive to race conditions. So the code has to have a certain simplicity at all places, such that developers do not have to think or wonder too much. We don't just need to write for the now, but for long-term maintenance. That 2% gain really doesn't matter if it leads to harder maintainability.

So now that we have the performance gains, let's focus on how to implement this as simple as possible; in particular the two transformation classes, which I think should become one. Then every method has to become really simple, and essentially only contain the implementation of one idea.

asynciterator.ts
jeswr: My thinking was - if we just pass an array forwards then we need to create a copy of the array each time, whereas with the linked list the functions are only stored once. This means that we end up taking up a lot of unnecessary space if we are doing a large number of mappings

asynciterator.ts
jeswr: Yeah, I've tried to go down that path, but it is very necessary to have the break condition rather than just forwarding nulls as this can allow filtered operations to terminate much faster; that said, we can remove the labels by refactoring 1344-1347 into a separate function and use return whilst seeing similar perf. outcomes

jeswr:
> Thanks for your work! The ideas are definitely good, but a significant amount of polishing is required.

Of course!

> That 2% gain really doesn't matter if it leads to harder maintainability.

If it was indeed 2% I would agree, but its a matter of 2x (1sec vs. 0 sec to 500maps on 1million elements) in terms of using this indexed loop structure rather than 'compiling' the function. That said, we can just write the compiled function using this nested loop and see similar results.

> So now that we have the performance gains, let's focus on how to implement this as simple as possible; in particular the two transformation classes, which I think should become one. Then every method has to become really simple, and essentially only contain the implementation of one idea.


Sounds good :)

asynciterator.ts
RubenVerborgh: We can definitely do this with two regular loops. Quick cut 'n' paste draft:

```typescript
  read(): D | null {
    const { source, funcs } = this;
    let item;
    while ((item = source.read()) !== null) {
      let next = funcs.length;
      let transformed = item;
      while (transformed !== null) {
        // …
      }
      if (transformed !== null)
        return transformed;
    }
    return null;
  }
  ```

File: asynciterator.ts (hunk 14 above)
Comment (RubenVerborgh) [reply asynciterator.ts]: Well we are creating the array anyway in `this._funcs = []`.

**Comment by RubenVerborgh** [comment]
Thanks, @jeswr!

> If it was indeed 2% I would agree

It was just a fictitious 2% as an example, but I see the confusion I created 🙂

In general, I meant to say that: we've got the perf now. Let's look at approaching that same perf with the clearest code possible—and I don't mind sacrificing a couple of percentages for long-term maintainability. That is: unless there is a report of an actual performance _problem_, we should not prematurely optimize. For instance, 500 maps are a case I've never seen in practice; and whereas it is good to scale, scaling in such a high number of transformations has so far not been necessary.

File: asynciterator.ts (hunk 13 above)
Comment (jeswr) [reply asynciterator.ts]: I've pushed an alternative that I had worked on before; but happy to look into this one further if you think the approach I now have up is still confusing

File: asynciterator.ts (hunk 14 above)
Comment (jeswr) [reply asynciterator.ts]: Ah but not in all the intermediary iterators, just one the ones that have `read` caelld on them. So we have just one length `n` array but not a length `n-1`, `n-2`, ..., `1`

File: asynciterator.ts (hunk 3 above)
Comment (jeswr) [reply asynciterator.ts:1335]: I don't think so without introducing a performance cost, as it is getting filtered out at

```ts
for (let index = endIndex; index >= 1; index -= 1) {
          if ((item = funcs[index](item)) === null)
            return null;
}
```

asynciterator.ts
jeswr: The source mechanics from the `TransformIterator` are a bit heavyweight esp since it extends from the `BufferedIterator`.

jeswr:
> For instance, 500 maps are a case I've never seen in practice; 

Good point, and this number of maps was unnecessary to see non-trivial perf. gains, 

For the code below on DELL XPS15 32G ram we get the following results (@jacoscaz - I don't think you were using a large enough Array to see the full effect)
|code|time(ms)|speedup|
|-|-|-|
|this PR|223|21.2x|
|#57|308|15.4x|
|main|4736|1x|

```ts
import { ArrayIterator } from './dist/asynciterator.js'

let i = 0;
const arr = new Array(2000000).fill(true).map(() => i++);

const iterator = new ArrayIterator(arr)
  .map((item) => item)
  .map((item) => item)
  .map((item) => item)
  .map((item) => item)
  .map((item) => item)
  .map((item) => item)
  .map((item) => item)
  .map((item) => item)
  .map((item) => item)
  .map((item) => item);

const now = Date.now();
iterator.on('data', () => {}).on('end', () => {
});
```

RubenVerborgh:
Okay, we're onto something!

For benchmarks, perhaps rather than one 2000000-element iterator, create 100 20000-element iterators or so. Numbers might or might not be different.

Once we have decided on the implementation strategy (cfr. other PRs) and the code has matured, we can go ahead with this.

jacoscaz:
Here's a test that goes through 20_000 items 100 times, each time applying 2 `map()` and 1 `filter()` transformations.

```typescript
let i = 0;
const arr = new Array(20_000).fill(true).map(() => i++);

const now = Date.now();
let times = 100;

const loop = () => {
  if (times === 0) {
    return;
  }
  const iterator = new ArrayIterator(arr)
    .map((item: number) => item)
    .map((item: number) => item)
    .filter((item: number) => item % 2 === 0)
  ;
  iterator.on('data', () => {}).on('end', () => {
    times -= 1;
    loop();
  });
};

loop();
```

On my machine (MacBook Pro, Apple Silicon, M1, 16 GB) this takes ~770 ms with the current `main` and ~90 ms with this branch, so roughly 8.5 times faster.

jeswr:
How does https://github.com/RubenVerborgh/AsyncIterator/pull/59 compare on your machine @jacoscaz ?

jacoscaz:
@jeswr with the same test as my previous comment, #59 takes ~110 ms, so roughly 1.2 times slower than this branch.

A 20% difference over the lifespan of an engine like Comunica can have significant effects but I agree with @RubenVerborgh that there's a line after which maintainability must be favored over raw performance. I'd be happy with either branches, really.

jeswr:
I'd imagine that the 20% slowdown is either caused by https://github.com/RubenVerborgh/AsyncIterator/pull/59/commits/9e64c82e48367b259dae8dbb9e248c70071f74ce or https://github.com/RubenVerborgh/AsyncIterator/pull/59/commits/f03b5413d470af0ef2ecc24934625ab697727459; so I would prefer to still use the other branch and just revert by a commit or 2

@jacoscaz sorry to be a pain, but my machine is a bit too consumed with other tasks to do good perf. testing so are you able to work out which commit is the problem?

jacoscaz:
Perf at the last 3 commits of #59:

- 5bbc92b09fb7336adbba5135911042c1126b79d8 ~98 - 100 ms, 1.1x slower 
- f03b5413d470af0ef2ecc24934625ab697727459 ~100 - 105 ms, still roughly 1.1x slower
- 9e64c82e48367b259dae8dbb9e248c70071f74ce ~105 - 115 ms, getting closer to 1.2x slower

I've run the test 5 times per commit. While the results are pretty consistent when it comes to which commit goes faster, the relative difference between one commit and the next is not as consistent. With these latest numbers I would definitely go for whichever branch is more maintainable.

RubenVerborgh:
^ Let's not draw conclusions from 100ms though; we'll need tests that run a couple of seconds, preferably with a warm-up.

RubenVerborgh:
See #59.

asynciterator.ts
```diff
@@ -520,7 +527,7 @@ export class AsyncIterator<T> extends EventEmitter {
     @returns {module:asynciterator.AsyncIterator} A new iterator with at most the given number of items
   */
   take(limit: number): AsyncIterator<T> {
-    return this.transform({ limit });
+    return new HeadIterator(this, limit);
```
RubenVerborgh, line 530: A quick inline doodle:

```typescript
  take(limit: number): AsyncIterator<T> {
    const source = this;
    const guard = {} as any;
    return new class extends MappingIterator<T, T> {
      constructor() {
        super(source, item => limit-- > 0 ? item : guard);
      }

      read(): T | null {
        const item = super.read();
        if (item !== guard)
          return item;
        this.close();
        return null;
      }
    };
  }
```

asynciterator.ts
```diff
@@ -1251,6 +1258,119 @@ function destinationFillBuffer<S>(this: InternalSource<S>) {
...
+
+export class HeadIterator<T> extends MappingIterator<T> {
+  protected count: number = 0;
+
+  constructor(source: AsyncIterator<T>, protected readonly limit: number) {
+    super(source);
+  }
+
+  read(): T | null {
+    const item = this.source.read();
```
RubenVerborgh, line 1365: Should probably be `super.read()` so all checks (are we not closed?) are still done.

PR #63, opened 2022-04-07
feat: faster wrapping

jacoscaz:
This PR covers step 2. in the plan described in https://github.com/RubenVerborgh/AsyncIterator/issues/44#issuecomment-1086950726 and aims to 1) lower the overhead of the `wrap()` function and 2) expand support to additional types of sources (namely `Iterator` and `Iterable` objects).

The rationale for this PR is taking the onus of figuring out how to best wrap common types of data sources away from users of this library. Users should be able to expect `wrap()` to use the strategy that is most appropriate for any given source (as long as it's a supported source).

Performance-wise, this PR does two things:

1) It introduces source-specific iterators that synchronously read from their sources, skipping the internal buffering done by `TransformIterator` (i.e. the class currently used for wrapping)

2) It introduces a specific option that allows `AsyncIterator` instances to fall-through as they are, with no wrapping applied

The following test, admittedly naive, takes ~116 ms with the current `main` branch and ~8 ms with this PR (`letIteratorThrough` makes virtually no difference). This mostly highlight the difference made by skipping `TransformIterator`'s internal buffering.

```typescript
let i = 0;
const arr = new Array(200_000).fill(true).map(() => i++);
const source = new ArrayIterator(arr);
const opts = { letIteratorThrough: false };
const then = Date.now();
wrap(source, opts).on('data', () => {}).on('end', () => {
  console.log('elapsed', Date.now() - then, 'ms');
});
```

Things become much more interesting when we consider what happens with a chain established by `wrap()`ing:

```typescript
let i = 0;
const arr = new Array(200_000).fill(true).map(() => i++);
const source = new ArrayIterator(arr);
const opts = { letIteratorThrough: false };
const then = Date.now();
wrap(wrap(wrap(wrap(source, opts), opts), opts), opts).on('data', () => {}).on('end', () => {
  console.log('elapsed', Date.now() - then, 'ms');
});
```

The above test takes ~250 ms with the current `main` branch, ~ 15 ms with this PR and `letIteratorThrough: false`, ~ 8 ms with this PR and `letIteratorThrough: true`. I haven't seen consistent differences with or without warmups.

My use case is the optimization of the interface between [quadstore](https://github.com/belayeng/quadstore), which implements the RDF/JS `Source` interface, and [comunica](https://comunica.dev), through which the former can be queried via SPARQL: https://github.com/comunica/comunica/blob/a2f31270778730911bd0bdf8b8c7ade567153319/packages/actor-rdf-resolve-quad-pattern-rdfjs-source/lib/RdfJsQuadSource.ts#L29 .

As many algebraic operation (matching, joining, ...) are happening in-memory, thus requiring the streaming of large amounts of quads from the persistence layer into the engine, removing even just one layer of indirection/buffering can make a significant difference.

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -2050,11 +2050,11 @@ export class WrappingIterator<T> extends AsyncIterator<T> {
 
     // Enable reading from source
     this._source = source;
-    this.readable = true;
+    this.readable = source.readable !== false;
   }
 
   read(): T | null {
-    if (this._source !== null) {
+    if (this._source !== null && this._source.readable !== false) {
       const item = this._source.read();
       if (item !== null)
         return item;
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -25,10 +25,6 @@ export function setTaskScheduler(scheduler: TaskScheduler): void {
   taskScheduler = scheduler;
 }
 
-// Returns a function that calls `fn` with `self` as `this` pointer. */
-function bind<T extends Function>(fn: T, self?: object): T {
-  return self ? fn.bind(self) : fn;
-}
 
 /**
   ID of the INIT state.
@@ -868,7 +864,7 @@ export class MappingIterator<S, D = S> extends AsyncIterator<D> {
 // Validates an AsyncIterator for use as a source within another AsyncIterator
 function ensureSourceAvailable<S>(source?: AsyncIterator<S>, allowDestination = false) {
   if (!source || !isFunction(source.read) || !isFunction(source.on))
-    throw new Error(`Invalid source: ${source}`);
+    throw new TypeError(`Invalid source: ${source}`);
   if (!allowDestination && (source as any)._destination)
     throw new Error('The source already has a destination');
   return source as InternalSource<S>;
@@ -1175,8 +1171,7 @@ export class TransformIterator<S, D = S> extends BufferedIterator<D> {
     @param {module:asynciterator.AsyncIterator} [options.source] The source this iterator generates items from
   */
   constructor(source?: SourceExpression<S>,
-              options: TransformIteratorOptions<S> =
-                source as TransformIteratorOptions<S> || {}) {
+              options: TransformIteratorOptions<S> = source as TransformIteratorOptions<S> || {}) {
     super(options);
 
     // Shift parameters if needed
@@ -1982,22 +1977,124 @@ class HistoryReader<T> {
 }
 
 /**
+ * An iterator that takes a variety of iterable objects as a source.
+ */
+export class WrappingIterator<T> extends AsyncIterator<T> {
+  protected _source: InternalSource<T> | null = null;
+
+  constructor(source: MaybePromise<IterableSource<T>>) {
+    super();
+    if (!isPromise(source))
+      this.source = source as any;
+    else
+      source.then(s => this.source = s as any).catch(e => this.emit('error', e));
+  }
+
+  protected set source(source: InternalSource<T>) {
+    // Process an iterable source
+    if (isIterable(source))
+      source = source[Symbol.iterator]() as any;
+    // Process an iterator source
+    if (isIterator<T>(source)) {
+      let iterator: Iterator<T> | null = source;
+      source = new EventEmitter() as any;
+      source.read = (): T | null => {
+        if (iterator !== null) {
+          const item = iterator.next();
+          if (!item.done)
+            return item.value;
+          // No remaining values, so stop iterating
+          iterator = null;
+          this.close();
+        }
+        return null;
+      };
+    }
+    // Process any other readable source
+    else {
+      source = ensureSourceAvailable(source);
+    }
+
+    // Set up event handling
+    source._destination = this;
+    source.on('end', destinationClose);
+    source.on('error', destinationEmitError);
+    source.on('readable', destinationSetReadable);
+
+    // Enable reading from source
+    this._source = source;
+    this.readable = true;
+  }
+
+  read(): T | null {
+    if (this._source !== null) {
+      const item = this._source.read();
+      if (item !== null)
+        return item;
+      this.readable = false;
+    }
+    return null;
+  }
+
+  protected _end(destroy: boolean = false) {
+    super._end(destroy);
+    // Clean up event handlers
+    if (this._source !== null) {
+      this._source.removeListener('end', destinationClose);
+      this._source.removeListener('error', destinationEmitError);
+      this._source.removeListener('readable', destinationSetReadable);
+      delete this._source._destination;
+      this._source = null;
+    }
+  }
+}
+
+
+/**
   Creates an iterator that wraps around a given iterator or readable stream.
   Use this to convert an iterator-like object into a full-featured AsyncIterator.
   After this operation, only read the returned iterator instead of the given one.
   @function
-  @param {module:asynciterator.AsyncIterator|Readable} [source] The source this iterator generates items from
+  @param [source] The source this iterator generates items from
   @param {object} [options] Settings of the iterator
   @returns {module:asynciterator.AsyncIterator} A new iterator with the items from the given iterator
 */
-export function wrap<T>(source: EventEmitter | Promise<EventEmitter>, options?: TransformIteratorOptions<T>) {
-  return new TransformIterator<T>(source as AsyncIterator<T> | Promise<AsyncIterator<T>>, options);
+export function wrap<T>(source: null | undefined): AsyncIterator<T>;
+export function wrap<T>(source: MaybePromise<IterableSource<T>>): AsyncIterator<T>;
+export function wrap<T>(source: MaybePromise<AsyncIterator<T>>,
+                        options?: TransformIteratorOptions<T>): AsyncIterator<T>;
+export function wrap<T>(source?: MaybePromise<IterableSource<T>> | null,
+                        options?: TransformIteratorOptions<T>,): AsyncIterator<T> {
+  // TransformIterator if TransformIteratorOptions were specified
+  if (options)
+    return new TransformIterator<T>(source as MaybePromise<AsyncIterator<T>>, options);
+
+  // Empty iterator if no source specified
+  if (!source)
+    return empty();
+
+  // Unwrap promised sources
+  if (isPromise<T>(source))
+    return new WrappingIterator(source);
+
+  // Directly return any AsyncIterator
+  if (source instanceof AsyncIterator)
+    return source;
+
+  // Other iterable objects
+  if (Array.isArray(source))
+    return fromArray<T>(source);
+  if (isIterable(source) || isIterator(source) || isEventEmitter(source))
+    return new WrappingIterator<T>(source);
+
+  // Other types are unsupported
+  throw new TypeError(`Invalid source: ${source}`);
 }
 
 /**
   Creates an empty iterator.
  */
-export function empty<T>() {
+export function empty<T>(): AsyncIterator<T> {
   return new EmptyIterator<T>();
 }
 
@@ -2005,7 +2102,7 @@ export function empty<T>() {
   Creates an iterator with a single item.
   @param {object} item the item
  */
-export function single<T>(item: T) {
+export function single<T>(item: T): AsyncIterator<T> {
   return new SingletonIterator<T>(item);
 }
 
@@ -2013,8 +2110,24 @@ export function single<T>(item: T) {
   Creates an iterator for the given array.
   @param {Array} items the items
  */
-export function fromArray<T>(items: Iterable<T>) {
+export function fromArray<T>(items: Iterable<T>): AsyncIterator<T> {
   return new ArrayIterator<T>(items);
+}
+
+/**
+ Creates an iterator for the given Iterator.
+ @param {Iterable} source the iterator
+ */
+export function fromIterator<T>(source: Iterable<T> | Iterator<T>): AsyncIterator<T> {
+  return new WrappingIterator<T>(source);
+}
+
+/**
+ Creates an iterator for the given Iterable.
+ @param {Iterable} source the iterable
+ */
+export function fromIterable<T>(source: Iterable<T> | Iterator<T>): AsyncIterator<T> {
+  return new WrappingIterator<T>(source);
 }
 
 /**
@@ -2035,25 +2148,12 @@ export function range(start: number, end: number, step?: number) {
   return new IntegerIterator({ start, end, step });
 }
 
-// Determines whether the given object is a function
-function isFunction(object: any): object is Function {
-  return typeof object === 'function';
-}
-
-// Determines whether the given object is an EventEmitter
-function isEventEmitter(object: any): object is EventEmitter {
-  return object && typeof object.on === 'function';
-}
-
-// Determines whether the given object is a promise
-function isPromise<T>(object: any): object is Promise<T> {
-  return object && typeof object.then === 'function';
-}
-
-// Determines whether the given object is a source expression
-function isSourceExpression<T>(object: any): object is SourceExpression<T> {
-  return object && (isEventEmitter(object) || isPromise(object) || isFunction(object));
-}
+export type IterableSource<T> =
+  T[] |
+  AsyncIterator<T> |
+  EventEmitter |
+  Iterator<T> |
+  Iterable<T>;
 
 export interface SourcedIteratorOptions {
   destroySource?: boolean;
@@ -2098,3 +2198,38 @@ type SourceExpression<T> =
 
 type InternalSource<T> =
   AsyncIterator<T> & { _destination?: AsyncIterator<any> };
+
+// Returns a function that calls `fn` with `self` as `this` pointer. */
+function bind<T extends Function>(fn: T, self?: object): T {
+  return self ? fn.bind(self) : fn;
+}
+
+// Determines whether the given object is a function
+export function isFunction(object: any): object is Function {
+  return typeof object === 'function';
+}
+
+// Determines whether the given object is an EventEmitter
+export function isEventEmitter(object: any): object is EventEmitter {
+  return isFunction(object?.on);
+}
+
+// Determines whether the given object is a promise
+export function isPromise<T>(object: any): object is Promise<T> {
+  return isFunction(object?.then);
+}
+
+// Determines whether the given object is a source expression
+export function isSourceExpression<T>(object: any): object is SourceExpression<T> {
+  return object && (isEventEmitter(object) || isPromise(object) || isFunction(object));
+}
+
+// Determines whether the given object supports the iterable protocol
+export function isIterable<T>(object: { [key: string]: any }): object is Iterable<T> {
+  return object && (Symbol.iterator in object);
+}
+
+// Determines whether the given object supports the iterator protocol
+export function isIterator<T>(object: { [key: string]: any }): object is Iterator<T> {
+  return isFunction(object?.next);
+}
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -1990,18 +1990,30 @@ class HistoryReader<T> {
 export class WrappingIterator<T> extends AsyncIterator<T> {
   protected _source: InternalSource<T> | null = null;
 
-  constructor(source: MaybePromise<IterableSource<T>>) {
+  constructor(source?: MaybePromise<IterableSource<T>>) {
     super();
-    if (!isPromise(source))
-      this.source = source as any;
-    else
-      source.then(s => this.source = s as any).catch(e => this.emit('error', e));
+
+    // If promise, set up a temporary source and replace when ready
+    if (isPromise(source)) {
+      this._source = new AsyncIterator() as any;
+      source.then(value => {
+        this._source = null;
+        this.source = value;
+      }).catch(error => this.emit('error', error));
+    }
+    // Otherwise, set the source synchronously
+    else if (source) {
+      this.source = source;
+    }
   }
 
-  protected set source(source: InternalSource<T>) {
+  set source(value: IterableSource<T>) {
+    let source: InternalSource<T> = value as any;
     // Do not change sources if the iterator is already done
     if (this.done)
       return;
+    if (this._source !== null)
+      throw new Error('The source cannot be changed after it has been set');
 
     // Process an iterable source
     if (isIterable(source))
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -2000,9 +2000,12 @@ export class WrappingIterator<T> extends AsyncIterator<T> {
       source = new EventEmitter() as any;
       source.read = (): T | null => {
         if (iterator !== null) {
-          const item = iterator.next();
-          if (!item.done)
-            return item.value;
+          // Skip any null values inside of the iterator
+          let next: IteratorResult<T>;
+          while (!(next = iterator.next()).done) {
+            if (next.value !== null)
+              return next.value;
+          }
           // No remaining values, so stop iterating
           iterator = null;
           this.close();
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -320,7 +320,7 @@ export class AsyncIterator<T> extends EventEmitter {
     const items: T[] = [];
     const limit = typeof options?.limit === 'number' ? options.limit : Infinity;
 
-    return limit <= 0 ? Promise.resolve(items) : new Promise<T[]>((resolve, reject) => {
+    return this.ended || limit <= 0 ? Promise.resolve(items) : new Promise<T[]>((resolve, reject) => {
       // Collect and return all items up to the limit
       const resolveItems = () => resolve(items);
       const pushItem = (item: T) => {
@@ -1203,6 +1203,10 @@ export class TransformIterator<S, D = S> extends BufferedIterator<D> {
   }
 
   set source(value: AsyncIterator<S> | undefined) {
+    // Do not change sources if the iterator is already done
+    if (this.done)
+      return;
+
     // Validate and set source
     const source = this._source = this._validateSource(value);
     source._destination = this;
@@ -1784,6 +1788,10 @@ export class ClonedIterator<T> extends TransformIterator<T> {
   }
 
   set source(value: AsyncIterator<T> | undefined) {
+    // Do not change sources if the iterator is already done
+    if (this.done)
+      return;
+
     // Validate and set the source
     const source = this._source = this._validateSource(value);
     // Create a history reader for the source if none already existed
@@ -1991,6 +1999,10 @@ export class WrappingIterator<T> extends AsyncIterator<T> {
   }
 
   protected set source(source: InternalSource<T>) {
+    // Do not change sources if the iterator is already done
+    if (this.done)
+      return;
+
     // Process an iterable source
     if (isIterable(source))
       source = source[Symbol.iterator]() as any;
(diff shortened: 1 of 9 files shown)


jacoscaz:
I should add that this PR is not ready to be merged as it's lacking tests and I'm sure the code can be improved. However, it is ready for a first round of feedbacks to better direct my efforts. Tagging @jeswr , @rubensworks and @RubenVerborgh to keep everyone in the loop. BTW, this would have to be rebased on `main` once #59 gets merged.

asynciterator.ts
```diff
@@ -1852,17 +1889,167 @@ class HistoryReader<T> {
...
+      this.emit('readable');
+    else
+      this.readable = true;
+  }
+
+  protected _onSourceEnd() {
+    this.close();
+    this._cleanup();
+  }
+
```
jeswr: You can follow the same logic as [here](https://github.com/RubenVerborgh/AsyncIterator/blob/6b4f039d5d12dc48256695f4133c70a931a490b7/asynciterator.ts#L1255-L1264) and move these to being functions outside the class. Put the `_cleanup` logic in `close` and just [reflect the readable status of the source](https://github.com/RubenVerborgh/AsyncIterator/blob/6b4f039d5d12dc48256695f4133c70a931a490b7/asynciterator.ts#L1269-L1275)

asynciterator.ts
```diff
@@ -1852,17 +1889,167 @@ class HistoryReader<T> {
...
+    return fromArray<T>(source);
+  if (isAsyncIteratorLike<T>(source))
+    return fromAsyncIteratorLike<T>(source);
+  if (!options.prioritizeIterable) {
+    if (isIterator<T>(source))
+      return fromIterator<T>(source);
+    if (isIterable<T>(source))
+      return fromIterable<T>(source);
+  }
+  throw new Error(`Unsupported source ${source}`);
```
jeswr: Refactor common logic (e.g. isIterator checks) into a helper functions

jeswr:
> Things become much more interesting when we consider what happens with a chain established by wrap()ing:

I've also seen this pattern in the core of Comunica (with a few synchronous transforms in between), mainly in the context of handling iterators that are wrapped in a promise. It would be good to have benchmarks on that case as well - i.e.

```ts
let i = 0;
const arr = new Array(200_000).fill(true).map(() => i++);
const source = new ArrayIterator(arr);
const opts = { letIteratorThrough: false };
const then = Date.now();
wrap(Promise.resolve(wrap(Promise.resolve(wrap(Promise.resolve(wrap(Promise.resolve(source), opts)), opts)), opts)), opts).on('data', () => {}).on('end', () => {
  console.log('elapsed', Date.now() - then, 'ms');
});
```

jacoscaz:
@jeswr I've tweaked your test a bit to ignore the time spent establishing the chain, just in case it could make a difference (likely insignificant, but still).

```typescript
let i = 0;
const arr = new Array(200_000).fill(true).map(() => i++);

const source = new ArrayIterator(arr);
const opts = { /* letIteratorThrough: true */ };

const chain = wrap(Promise.resolve(wrap(Promise.resolve(wrap(Promise.resolve(wrap(Promise.resolve(source), opts)), opts)), opts)), opts);

chain.once('readable', () => {
  const then = Date.now();
  chain.on('data', () => {}).on('end', () => {
    console.log('elapsed', Date.now() - then, 'ms');
  });
});
```

This test runs in ~450 ms with current `main` and ~12 ms with this PR. `letIteratorThrough` makes no difference as having to return an iterator _synchronously_ forces the use of `WrappingIterator` even for `AsyncIterator` sources as we still need to wait for the promise to asynchronously resolve.

asynciterator.ts
jacoscaz: I can move those functions outside of the `WrappingIterator` class but I can't reflect the readable status of the source as I cannot assume the latter to be an instance of `AsyncIterator` and/or have a boolean `readable` property. I notice that `stream.Readable` has had that property since Node.js `v11.4.0` but this package supports versions all the way back to `v10`. And even then, the [RDF/JS `Stream` interface](https://rdf.js.org/stream-spec/#stream-interface) does not mandate a `readable` property.

jacoscaz: I also can't add the `_destination` property to `WrappingIterator#_source` as the latter can be user-provided (if we detect it to be an `AsyncIteratorLike`)... I'll add a couple of dedicated functions.

asynciterator.ts
jacoscaz: Done, reopen if needed.

asynciterator.ts
jacoscaz: I ended up moving cleanup logic into `protected _end()` but I can't think of other clean ways to improve on the above:

1) I can't reflect the readable status of the source because it might not have one (RDF/JS streams, `stream.Readable` when using Node.js version `[10.0 - 11.4), ...`.
2) I can't add `_destination` to `WrappingIterator#_source` or modify the source in any way as it might be user-provided, so I can't use `destinationEmitError` and the likes.
3) I still need to keep references to event listeners around so that I can call `removeListener()` when the wrapping iterator is closed / destroyed.

Leaving this open in case you have suggestions @jeswr , otherwise feel free to resolve.

asynciterator.ts
```diff
@@ -25,6 +25,48 @@ export function setTaskScheduler(scheduler: TaskScheduler): void {
...
+// Determines whether the given object is an instance of ES Iterable
+export const isIterable = <T>(item: { [key: string]: any }): item is Iterable<T> => {
+  return Symbol.iterator in item;
+};
+
+// Determines whether the given object is an instance of ES Iterable or an instance of ES Iterator
+export const isIterableOrIterator = <T>(item: { [key: string]: any }): item is Iterable<T> | Iterator<T> => {
+  return isIterable<T>(item) || isIterator<T>(item);
+};
+
```
jeswr: Why are you using arrow-body style here?

asynciterator.ts
```diff
@@ -1852,17 +1894,166 @@ class HistoryReader<T> {
...
+    let value: T | null = null;
+    if (this._source) {
+      const item = this._source.next();
+      if (!item.done)
+        value = item.value;
+      else
+        this.close();
+    }
+    return value;
+  }
```
jeswr: So as to avoid unecessary variable assignment, and to avoid branching on the common case, why not

```ts
if (!this._source)
  return null
const item = this._source.next();
if (item.done) {
   this.close();
   return null;
}
return item.value
```

asynciterator.ts
```diff
@@ -1852,17 +1894,166 @@ class HistoryReader<T> {
...
+ * sources.
+ */
+export class WrappingIterator<T> extends AsyncIterator<T> {
+  protected _source?: AsyncIteratorLike<T>;
+
+  constructor(sourceOrPromise: WrapSource<T> | Promise<WrapSource<T>> | null | undefined, options: WrapOptions = {}) {
+    super();
+    this._onSourceEnd = this._onSourceEnd.bind(this);
+    this._onSourceError = this._onSourceError.bind(this);
+    this._onSourceReadable = this._onSourceReadable.bind(this);
```
jeswr: Avoid creating custom versions of these functions for each class - instead use globally defined functions like https://github.com/RubenVerborgh/AsyncIterator/blob/4652546f82cf6bbd949a61443a21ce57f40f12e8/asynciterator.ts#L850-L854.

In fact - I'd imagine that `iteratorIterator` and `MappingIterator` could share a common abstract class after you rebase.

asynciterator.ts
```diff
@@ -1852,17 +1894,166 @@ class HistoryReader<T> {
...
+      this._source.removeListener('error', this._onSourceError);
+      this._source.removeListener('readable', this._onSourceReadable);
+      delete this._source;
+    }
+  }
+
+  read(): T | null {
+    if (this._source)
+      return this._source.read();
+    return null;
```
jeswr: Why not use a ternary here?

asynciterator.ts
```diff
@@ -1852,17 +1894,166 @@ class HistoryReader<T> {
...
+  if (isPromise<T>(sourceOrPromise))
+    return new WrappingIterator(sourceOrPromise, options);
+  try {
+    return fromSource(sourceOrPromise as WrapSource<T>, options);
+  }
+  catch (err) {
+    const iterator = empty<T>();
+    scheduleTask(() => iterator.emit('error', err));
+    return iterator;
+  }
```
jeswr: This deviates from the way errors are specified in the readable spec https://nodejs.org/api/stream.html#event-error. Primarily, errors should only occur when *data* is being written or piped.

For the sake of future-proofing / compatibility with the `Readable` API - I would just throw an error instead rather than making it part of the stream.

test/WrappingIterator-test.js
```diff
@@ -0,0 +1,116 @@
...
+    it('should emit all items', async () => {
+      expect(await iterator.toArray()).to.deep.equal([0, 1, 2, 3, 4]);
+    });
+  });
+
+  describe('with a promisified stream.Readable source', () => {
+    let iterator;
+    let source;
+    before(() => {
+      source = Promise.resolve(Readable.from([0, 1, 2, 3, 4]));
```
jeswr, line 345: For the sake of checking you're handling events correctly please have a few more `Promise` tests with different source types, e.g. range iterators, actual arrays - and more importantly sources that will buffer.

asynciterator.ts
jacoscaz: Habit - I tend to favor arrow functions over normal functions. That said, other `is*` functions are normal functions so I changed the new ones into normal functions, too.

asynciterator.ts
jacoscaz: Done!

asynciterator.ts
jacoscaz: I find ternary iterators to be somewhat difficult to handle on the debugging side of things (as in when using a debugger). My first instinct is to use them, which is why you'll find them in other code bits of mine, but then I often come to regret it when debugging and thus I'm trying to force myself not to use them anymore.

asynciterator.ts
jacoscaz: I had been under the impression that this would have constituted a breaking change but, upon looking at `main` again, I realized I was wrong. Done!

asynciterator.ts
jacoscaz: I'm not sure as the sources we're dealing with in `IteratorIterator`, `MappingIterator` and `WrappingIterator` are of different kinds, with the latter class being the one where the source is less strictly-defined. Leaving this open and unaddressed until I rebase after #59 is merged.

test/WrappingIterator-test.js
jacoscaz: Done!

asynciterator.ts
jacoscaz: Closing in favor of https://github.com/RubenVerborgh/AsyncIterator/pull/63#discussion_r847747818 , which sort of includes this one.

jacoscaz:
Not sure about what's going on with coveralls but this is probably ready for review by @RubenVerborgh (in due time, after #59 is merged). I'll resume working on cloning.

jacoscaz:
@RubenVerborgh I have just rebased this one on `main`. I'm not sure as to what is happening with the failing test on Node 10.x but I would prefer if you were to take a look first before doing further work on this.

asynciterator.ts
```diff
@@ -2035,25 +2192,17 @@ export function range(start: number, end: number, step?: number) {
...
-function isEventEmitter(object: any): object is EventEmitter {
-  return object && typeof object.on === 'function';
-}
-
-// Determines whether the given object is a promise
-function isPromise<T>(object: any): object is Promise<T> {
-  return object && typeof object.then === 'function';
+export interface WrapOptions {
+  prioritizeIterable?: boolean;
+  letIteratorThrough?: boolean;
```
RubenVerborgh: @jacoscaz Could you explain what both options are doing?

I would try to make them obsolete (my preference) or rename them (if there is no way around having them).

jacoscaz: Hi @RubenVerborgh !

> @jacoscaz Could you explain what both options are doing?

- `prioritizeIterable` gives users more control over what interface might be used to iterate upon a source in case the source implements more than one of the supported interfaces. An array, for example, can be wrapped both via `fromArray` and via `fromIterable`.

- `letIteratorThrough` gives users the option of having `AsyncIterator` instances passed through as they are. I've done this a while ago so I'm not 100% sure but I think I added this mostly for backward compatibility.
 
> I would try to make them obsolete (my preference) or rename them (if there is no way around having them).

Looking at this again after a while, I would also make these obsolete and just have the library do whatever's best for the given source (including passing the latter through as-is). @jeswr might want to pitch in as he initially suggested `prioritizeIterable`.

RubenVerborgh:
Thanks, removed the second option 16fcfb638a97df3b26cb11aa3d4da060b120c146 and will talk to Jesse over lunch about the other one!

jacoscaz:
Awesome!

asynciterator.ts
RubenVerborgh: Thanks, removed both in e83a5104b3cb0078992a146beb75ed34e96f0813.

asynciterator.ts
```diff
@@ -1981,42 +1978,172 @@ class HistoryReader<T> {
...
+      this._source = isAsyncReadable(source) ? source : fromSource(source);
+    }
+    catch (error) {
+      scheduleTask(() => this.emit('error', error));
+      return;
+    }
+    this._source._destination = this;
+    this._source.on('end', destinationClose);
+    this._source.on('error', destinationEmitError);
+    this._source.on('readable', () => this.emit('readable'));
```
RubenVerborgh: @jacoscaz This line shouldn't be here, but there are errors when I remove it. Could you please have a look? (The relevant commit is 555d88e043797df8b9f4262779b13d7c4ed5734c)

jacoscaz: @RubenVerborgh Looking into it. As a more general consideration, I don't think using generic listeners here is a good idea. If `isAsyncReadable(source)` returns true, we could still be dealing with an externally-provided, async-ish object such as a `stream.Readable` and IMHO in that case we shouldn't add the `_destination` property to it. This is one of reasons why I had opted to use dedicated event listeners as opposed to the generic ones.

jacoscaz: Seems to be something related to the debouncing of the `readable` event that happens at https://github.com/RubenVerborgh/AsyncIterator/blob/555d88e043797df8b9f4262779b13d7c4ed5734c/asynciterator.ts#L253-L262 . Commenting the outer `if` out makes the two tests that are failing pass (whilst breaking ~50 others).

jacoscaz: This works:

```js
    it('should emit all items', async () => {
      return new Promise((resolve) => {
        const items = [];
        iterator.once('end', () => {
          expect(items).to.deep.equal([0, 1, 2, 3]);
          resolve();
        });
        // iterator.on('readable', () => {
          let item;
          while ((item = iterator.read()) !== null) {
            items.push(item);
            console.log('ITEM', item);
          }
        // });
    });
```

Whereas it doesn't work if I uncomment the `readable` listener registration.

RubenVerborgh: Perfect thanks; then it's solved indeed, the bug was in the test. The readable event is and should only be fired when a non-readable iterator becomes readable.

Regarding generic listeners: it's a trick I picked up from the Node.js source a long time ago; the idea was to minimize object creation in performance-critical code. However, this advice might or might not be outdated. In any case, the rest of AsyncIterator is doing it, so I will stick to it for consistency. The _destination approach could be changed in the next major version with a WeakMap, or indeed with new listener objects every time.

jeswr:
> Thanks, removed the second option https://github.com/RubenVerborgh/AsyncIterator/commit/16fcfb638a97df3b26cb11aa3d4da060b120c146 and will talk to Jesse over lunch about the other one!

The request for that was in relation to performance when wrapping N3.js.

By default `wrap(store.match(...))` would end up using `fromAsyncReadable` which in turn uses buffering within the `WrappingIterator` (and all results get buffered into an array as a result of [this](https://github.com/rdfjs/N3.js/blob/b2661908d8af3e57fcb67111b32752118a8590d0/src/N3Store.js#L795-L799)). As opposed to just using the `IterableWrapper` which is much faster as it just pulls out quads on-demand. IIRC we cannot change the default prioritization order for backwards compatibility reasons.

The main consequence of removing this is that downstream clients like Comunica will sometimes need to write 

```ts
 if ((isIterable(source) || isIterator(source)))
    return fromIterable<T>(source);
 else if
    return wrap<T>(source);
```

rather than
```ts
return wrap(source, { prioritizeIterable: true })
```

asynciterator.ts
```diff
@@ -8,6 +8,7 @@ import createTaskScheduler from './taskscheduler';
 import LinkedList from './linkedlist';
 import type { Task, TaskScheduler } from './taskscheduler';
 
+
```
jeswr:
```suggestion
```

Unecessary added break

asynciterator.ts
```diff
@@ -1981,42 +1978,172 @@ class HistoryReader<T> {
...
+        source.then(s => this._wrapSource(s as any)).catch(e => this.emit('error', e));
+    }
+  }
+
+  protected _wrapSource(source: InternalSource<T>) {
+    try {
+      this._source = isAsyncReadable(source) ? source : fromSource(source);
+    }
+    catch (error) {
+      scheduleTask(() => this.emit('error', error));
```
jeswr: Why does this need to be scheduled? (why can't this be called immediately?)

asynciterator.ts
jeswr:
```suggestion
```
Isn't this already covered by the `destinationSetReadable` line below since the `readable` event will be emitted when needed. If not I would reccomend creating a `destinationEmitReadable`.

This is also problematic since the listener cannot be removed when the iterator is closed.

RubenVerborgh:
Thanks @jeswr; I'll take a look at the backwards compat thing too then. But we should def put performance-friendly options first. (Also this will be one of the last changes before semver.major, so then we can happily break.)

asynciterator.ts
RubenVerborgh: Indeed; this was related to a bug in the tests which we've resolved.

asynciterator.ts
RubenVerborgh: Happens during construction; consumer has had no time to attach listener yet.

asynciterator.ts
jacoscaz: > Perfect thanks; then it's solved indeed, the bug was in the test. The readable event is and should only be fired when a non-readable iterator becomes readable.

Was it? the `toArray()` method should always work, regardless of whether it's invoked when the iterator is readable or not. I could very well be missing something but this looks still unaddressed to me.

jacoscaz:
> Thanks @jeswr; I'll take a look at the backwards compat thing too then. But we should def put performance-friendly options first. (Also this will be one of the last changes before semver.major, so then we can happily break.)

Thanks @jeswr for the refresher. I hadn't realized we already have a use case for the `prioritizeIterable` flag. Long-term, I suspect more such flags might follow (e.g. for [native async iteration](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_async_iterator_and_async_iterable_protocols)). Perhaps an explicit ordering might be better, here? Something like `priority: ['iterable', 'asynciterable', 'array', ...]`?

asynciterator.ts
jacoscaz: > so I will stick to it for consistency

Consistency is one of the many reasons why I like working on `asynciterator` so much. I made your [Programming is an art](https://ruben.verborgh.org/blog/2013/02/21/programming-is-an-art/) post mandatory reading in more than one occasion. Still, I don't think we should modify externally provided objects. Collisions of property names are unlikely but not impossible and if one such collision were to happen it would lead to an extremely hard-to-debug behavior.

jacoscaz: @RubenVerborgh found the culprit and fixed in https://github.com/RubenVerborgh/AsyncIterator/pull/63/commits/159ed2b09e599d03d574aec2c98b40ae416d168b . Basically, I was ignoring the fact that a `null` return by `WrappingIterator#read()` should trigger a switch to `readable === false`, so that subsequent `readable` events from the source can actually result in a `readable` event from the wrapper.

jacoscaz: > Collisions of property names are unlikely but not impossible and if one such collision were to happen it would lead to an extremely hard-to-debug behavior.

@RubenVerborgh thinking about it, a non-breaking workaround for this would be to rename the `_destination` property to a `Symbol()` (see the [relevant MDN page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol)). That way the chances of collision would become virtually non-existent. This would be out of scope for this PR and better addressed in a dedicated one, so if my fix for the missing readable event is good I think this thread can go back to being resolved.

jacoscaz: Gave it a go in https://github.com/RubenVerborgh/AsyncIterator/pull/78 .

RubenVerborgh: Perfect, thanks for both of these!

RubenVerborgh (approved):

RubenVerborgh:
@jacoscaz @jeswr Done my in-depth review—what do you think?

jacoscaz:
I like where this ended at! Thanks @RubenVerborgh , +1 from me to merge and I’ll try to rebase #78 post-merge as soon as possible so that it can also be merged, completing work on wrapping.

asynciterator.ts
```diff
@@ -1981,42 +1984,170 @@ class HistoryReader<T> {
...
+export class WrappingIterator<T> extends AsyncIterator<T> {
+  protected _source: InternalSource<T> | null = null;
+
+  constructor(source?: MaybePromise<IterableSource<T>>) {
+    super();
+    if (source) {
+      if (!isPromise(source))
+        this.source = source as any;
+      else
+        source.then(s => this.source = s).catch(e => this.emit('error', e));
```
jeswr: This has the potential to lead to a gnarly race condition.

In particular if I run
```ts
const iter = new WrappingIterator(new Promise(resolve => {   
   setTimeout(() => { resolve(range(1, 5)) }, 1000)
   }))
iter.source = range(10)
```

Then when we start reading from the iterator it will return `10, 11, ...` for the first second whilst the input source is still resolving; and then it will go `1, 2, 3, 4, 5` before closing.

That is to say, needs to be some sort of way to invalidate this setting operation if `set source` is called afterwards.

---

The easier option is to make `set source` internal and have this warning about it's usage that links to this comment. Then no code changes are required.

asynciterator.ts
```diff
@@ -1981,42 +1984,170 @@ class HistoryReader<T> {
...
+
+    // Set up event handling
+    source._destination = this;
+    source.on('end', destinationClose);
+    source.on('error', destinationEmitError);
+    source.on('readable', destinationSetReadable);
+
+    // Enable reading from source
+    this._source = source;
+    this.readable = true;
```
jeswr: I would prefer `this.readable = source.readable` for the sake of performance; that way we don't trigger reading flows downstream if the source does not have any elements available.

asynciterator.ts
```diff
@@ -1981,42 +1984,170 @@ class HistoryReader<T> {
...
+        if (iterator !== null) {
+          // Skip any null values inside of the iterator
+          let next: IteratorResult<T>;
+          while (!(next = iterator.next()).done) {
+            if (next.value !== null)
+              return next.value;
+          }
+          // No remaining values, so stop iterating
+          this.close();
+          iterator = null;
```
jeswr, line 2034: I don't think this does anything as the iterator variable is never read after this point

asynciterator.ts
```diff
@@ -1981,42 +1984,170 @@ class HistoryReader<T> {
...
-  return new TransformIterator<T>(source as AsyncIterator<T> | Promise<AsyncIterator<T>>, options);
+export function wrap<T>(source: null | undefined): AsyncIterator<T>;
+export function wrap<T>(source: MaybePromise<IterableSource<T>>): AsyncIterator<T>;
+export function wrap<T>(source: MaybePromise<AsyncIterator<T>>,
+                        options?: TransformIteratorOptions<T>): AsyncIterator<T>;
+export function wrap<T>(source?: MaybePromise<IterableSource<T>> | null,
+                        options?: TransformIteratorOptions<T>,): AsyncIterator<T> {
+  // Empty iterator if no source specified
+  if (!source)
+    return empty();
```
jeswr: This could lead to bugs when `options.autoStart` is set to false; because `empty()` always autostarts. For now perhaps just do `return new ArrayIterator([], { autoStart: options.autoStart })` as a work around which we can return to `return empty()` when #45 is merged.

jeswr (changes requested):
Did a deeper review than last time and picked up on some potential bugs / race conditions. Apologies for not identifying them in the last review.

asynciterator.ts
jacoscaz: Big +1 on this, `set source()` on `WrappingIterator` should not be part of the public API.

asynciterator.ts
jacoscaz: We can't guarantee that property to be there, though... Perhaps `this.readable = typeof source.readable === 'boolean' ? source.readable : true`

asynciterator.ts
RubenVerborgh: Ugh and I was so glad I remembered 56fc2b23bb74028cde06ad2704e3257e79b696ae 😅

Absolutely right, we need to guard against pending promises.

> set source() on WrappingIterator should not be part of the public API

Actually there's an interesting use case, which is why I did 98eb9cfb19d4640cdb8d19083e7dedf1758e39b7: that way, you can use `WrappingIterator` as a placeholder and then insert the actual iterator when it is ready. `TransformIterator` supports that same pattern.

So I'll fix the pending promise issue.

asynciterator.ts
RubenVerborgh: Perfect, I'll combine both of your answers.

jeswr: Are you sure @jacoscaz. The lines above rely on it having a readable event?

jacoscaz: @jeswr  when it comes to the native `stream` module, the `readable` property came with Node.js 11.x if I recall correctly. Up until then we only had the `readable` event. Consequently, there's plenty of old versions of the userland `readable-stream` package out there without it.

RubenVerborgh:
(Just fixing a small mistake in 88165c48d15972fb0dc27bd81d2149562014264f that causes CI failures.)

jeswr (approved):

RubenVerborgh:
A major thank you to @jacoscaz as well as @jeswr for your efforts and patience.

I realize it has taken me a lot of time to review and complete this PR, so I definitely wanted to emphasize my appreciation for helping out with open source.

Because of its precise and high-performance nature, AsyncIterator is a special project in the sense that every PR tends to require one uninterrupted day for me to work on it. There's a lot of thinking work on finding which code makes it easiest to reason about correct behavior in a wide array of edge cases. So if you find me replacing or rewriting large chunks of code you have contributed, it's not you—it's me! Basically the hard work is finding the initial code that works, and covering it with an appropriate set of extensive tests, which is what you've done. Then it's my job to re-interpret the code that satisfies that those tests, and that's what tends to take me quite some time to get right.

Thanks for bearing with me, and I look forward to getting to the next PRs!

RubenVerborgh:
Published as v3.6.0 🥳

jacoscaz:
> Because of its precise and high-performance nature, AsyncIterator is a special project in the sense that every PR tends to require one uninterrupted day for me to work on it.

@RubenVerborgh thank _you_ for this library in the first place and all the effort that goes into maintaining it! I enjoy your reviews as they always lead to objectively better code from a correctness /  performance standpoint. I also enjoy seeing the same code rewritten in a style that is not the one I started out with - there's always a great deal to learn.

PR #64, opened 2022-04-08
feat: add uniq method to AsyncIterator

jeswr:
This add the method .uniq the the AsyncIterator which captures a common pattern used with AsyncIterators of filtering elements in the iterator such that only unique elements are passed through.

Since this uses the .filter method directly this will be unaffected by #59

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -479,6 +479,24 @@ export class AsyncIterator<T> extends EventEmitter {
   }
 
   /**
+   * Returns a new iterator containing all of the unique items in the original iterator.
+   * @param by - The derived value by which to determine uniqueness (e.g., stringification).
+                 Defaults to the identity function.
+   * @returns An iterator with duplicates filtered out.
+   */
+  uniq(by: (item: T) => any = identity): AsyncIterator<T> {
+    const uniques = new Set();
+    return this.filter(function (this: AsyncIterator<T>, item) {
+      const hashed = by.call(this, item);
+      if (!uniques.has(hashed)) {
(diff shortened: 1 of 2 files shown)


asynciterator.ts
```diff
@@ -472,6 +472,23 @@ export class AsyncIterator<T> extends EventEmitter {
     return this.transform({ filter: self ? filter.bind(self) : filter });
   }
 
+  /**
+   * Returns a new iterator containing all of the unique items in the original iterator.
+   * @param by The property by which to determine uniqueness (e.g. stringification). This is the identity
+   * by default.
+   * @returns An iterator with duplicates filtered out.
+   */
+  uniq(by: (item: T) => any = item => item): AsyncIterator<T> {
```
jacoscaz: I think the signature for `by` should read `(item: T) => string` to prevent serialization mistakes by passing functions that may lead to collisions due to implicit `toString()` calls resulting in hashes similar to `[Object object]`.

jacoscaz, line 481: I think the signature for `by` should read `(item: T) => string` to prevent serialization mistakes by passing functions that may lead to collisions due to implicit `toString()` calls resulting in hashes similar to `[Object object]`.

jacoscaz: Oh nevermind, I just realized `Set` works any kind of type. Silly me.

jacoscaz: Oh nevermind, I just realized `Set` works any kind of type. Silly me.

RubenVerborgh:
Thanks a lot—releasing this now as v3.7.0.

PR #65, opened 2022-04-10
WIP: V3 Sync Unions

### LINKED ISSUE #67: Performance: Improve performance of `append` and `prepend`
See use case for `append` in #66 .

`append` and `prepend` still go through the `SimpleTransformIterator`

IMO the cleanest way to do this would be to add a parameter to the `UnionIterator` in #65 that requires the iterators in the union to be consumed sequentially, and then for `append` and `prepend` to just return a `union` of the current source, and the data to `append`/`prepend`.

@jacoscaz - do you have any thoughts on this?

jeswr:
Same idea as #54 but targeted for v3.x.

#63 and #59 should be merged first

Now also resolves #67

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -479,7 +479,10 @@ export class AsyncIterator<T> extends EventEmitter {
     @returns {module:asynciterator.AsyncIterator} A new iterator that prepends items to this iterator
   */
   prepend(items: T[] | AsyncIterator<T>): AsyncIterator<T> {
-    return this.transform({ prepend: items });
+    return new UnionIterator(
+      [Array.isArray(items) ? new ArrayIterator(items, { autoStart: false }) : items, this],
+      { maxBufferSize: 1 }
+    );
   }
 
   /**
@@ -489,7 +492,10 @@ export class AsyncIterator<T> extends EventEmitter {
     @returns {module:asynciterator.AsyncIterator} A new iterator that appends items to this iterator
   */
   append(items: T[] | AsyncIterator<T>): AsyncIterator<T> {
-    return this.transform({ append: items });
+    return new UnionIterator(
+      [this, Array.isArray(items) ? new ArrayIterator(items, { autoStart: false }) : items],
+      { maxBufferSize: 1 }
+    );
   }
 
   /**
@@ -1583,6 +1589,8 @@ export class UnionIterator<T> extends AsyncIterator<T> {
     });
     if (this.buffer.empty && this._sources.done && this._sourceStarted)
       this.close();
+    else
+      this.readable = true;
   }
 
   public close() {
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -1519,6 +1519,7 @@ export class UnionIterator<T> extends AsyncIterator<T> {
   private _sources : AsyncIterator<AsyncIterator<T>>;
   private buffer = new LinkedList<AsyncIterator<T>>();
   private _sourceStarted: boolean;
+  private _maxBufferSize: number;
 
   /**
     Creates a new `UnionIterator`.
@@ -1549,6 +1550,7 @@ export class UnionIterator<T> extends AsyncIterator<T> {
     this._addSource(this._sources);
 
     this._sourceStarted = options.autoStart !== false;
+    this._maxBufferSize = options.maxBufferSize || Infinity;
     if (this._sources.done && this._sourceStarted)
       this.close();
     else
@@ -1572,7 +1574,7 @@ export class UnionIterator<T> extends AsyncIterator<T> {
   }
 
   protected _removeEmptySources() {
-    this.buffer.filter((source: any) => {
+    this.buffer.mutateFilter((source: any) => {
       if (source.done) {
         this._removeSource(source);
         return false;
@@ -1597,19 +1599,22 @@ export class UnionIterator<T> extends AsyncIterator<T> {
     const { buffer, _sources } = this;
     let item: T | null;
     let iterator: AsyncIterator<T> | null;
-    for (iterator of buffer) {
-      if (iterator.readable && (item = iterator.read()) !== null)
+    let node = buffer._head;
+    while (node !== null) {
+      if (node.value.readable && (item = node.value.read()) !== null)
         return item;
+      node = node.next;
     }
-    while ((iterator = _sources.read()) !== null) {
+
+    while (this.buffer.length < this._maxBufferSize && (iterator = _sources.read()) !== null) {
       this._addSource(iterator as any);
       this.buffer.push(iterator);
 
       if ((item = iterator.read()) !== null)
         return item;
     }
-    this.readable = false;
     this._removeEmptySources();
+    this.readable = false;
     return null;
   }
 }
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -1243,6 +1243,9 @@ export class TransformIterator<S, D = S> extends BufferedIterator<D> {
 function destinationEmitError<S>(this: InternalSource<S>, error: Error) {
   this._destination.emit('error', error);
 }
+function destinationSetReadable<S>(this: InternalSource<S>) {
+  this._destination.readable = true;
+}
 function destinationCloseWhenDone<S>(this: InternalSource<S>) {
   (this._destination as any)._closeWhenDone();
 }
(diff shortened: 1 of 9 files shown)


linkedlist.ts
```diff
@@ -34,8 +34,42 @@ export default class LinkedList<V> {
     return value;
   }
 
+  filter(filter: (item: V) => boolean) {
```
jacoscaz: This mutates the instance it is called upon. I would suggest keeping this aligned with `Array#filter()`, lodash/underscore's `.filter()` and most implementation of the same concept (insofar as my experience goes) by making it return a new instance. Or, renaming to something different that makes it clear we're mutating the original one.

linkedlist.ts
```diff
@@ -34,8 +34,42 @@ export default class LinkedList<V> {
...
+    }
+    this._tail = last;
+  }
+
   clear() {
     this._length = 0;
     this._head = this._tail = null;
   }
+
+  *[Symbol.iterator]() {
```
jacoscaz, line 70: We should clarify that this iterator does not keep yielding items as they are pushed into the list. It synchronously runs until the current end of the list and that's it.

I'm only pointing this out because I'm working on a cursor-like object that will never actually end, continuously moving to new items as they are added to the list, as the backing for the revised version of `HistoryReader`. We should be very explicit about the difference between these two.

jacoscaz, line 70: We should clarify that this iterator does not keep yielding items as they are pushed into the list. It synchronously runs until the current end of the list and that's it.

I'm only pointing this out because I'm working on a cursor-like object that will never actually end, continuously moving to new items as they are added to the list, as the backing for the revised version of `HistoryReader`. We should be very explicit about the difference between these two.

asynciterator.ts
```diff
@@ -1526,101 +1529,88 @@ export class UnionIterator<T> extends BufferedIterator<T> {
...
-        }
-      }
+  public read(): T | null {
+    if (!this._sourceStarted)
+      this._sourceStarted = true;
+
+    const { buffer, _sources } = this;
+    let item: T | null;
+    let iterator: AsyncIterator<T> | null;
+    for (iterator of buffer) {
```
jacoscaz: Perhaps a `forEach()` method could be more performant than the `Iterable` protocol?

jacoscaz: Perhaps a `forEach()` method could be more performant than the `Iterable` protocol?

asynciterator.ts
```diff
@@ -1526,101 +1529,88 @@ export class UnionIterator<T> extends BufferedIterator<T> {
...
+      if (iterator.readable && (item = iterator.read()) !== null)
+        return item;
     }
+    while ((iterator = _sources.read()) !== null) {
```
jacoscaz: If I understand correctly, we get here when all of the sources that we are currently reading from fail to return an item. If that happens, we add as many new sources as we can.

Can we add a limit to the number of sources we can be reading from at any one time? That would come useful in use cases where a source is backed by a stateful form of I/O with allocated resources (i.e. never good to have too many open connections or file descriptors).

jacoscaz: If I understand correctly, we get here when all of the sources that we are currently reading from fail to return an item. If that happens, we add as many new sources as we can.

Can we add a limit to the number of sources we can be reading from at any one time? That would come useful in use cases where a source is backed by a stateful form of I/O with allocated resources (i.e. never good to have too many open connections or file descriptors).

asynciterator.ts
jeswr: Good catch - it slows [[this example]](https://github.com/RubenVerborgh/AsyncIterator/pull/54#issue-1182327786) from 100ms to 300ms

jeswr:
Note: the failing test is because `.map` is buffering. This will be resolved once we rebase to https://github.com/RubenVerborgh/AsyncIterator/pull/59

asynciterator.ts
```diff
@@ -479,7 +479,10 @@ export class AsyncIterator<T> extends EventEmitter {
     @returns {module:asynciterator.AsyncIterator} A new iterator that prepends items to this iterator
   */
   prepend(items: T[] | AsyncIterator<T>): AsyncIterator<T> {
-    return this.transform({ prepend: items });
+    return new UnionIterator(
+      [Array.isArray(items) ? new ArrayIterator(items, { autoStart: false }) : items, this],
```
jacoscaz, line 483: After #63 is merged, this could become simply `wrap(items, { letIteratorThrough: true })` (thus supporting a wider range of types for `items`).

asynciterator.ts
```diff
@@ -489,7 +492,10 @@ export class AsyncIterator<T> extends EventEmitter {
     @returns {module:asynciterator.AsyncIterator} A new iterator that appends items to this iterator
   */
   append(items: T[] | AsyncIterator<T>): AsyncIterator<T> {
-    return this.transform({ append: items });
+    return new UnionIterator(
+      [this, Array.isArray(items) ? new ArrayIterator(items, { autoStart: false }) : items],
```
jacoscaz, line 496: Same as above

asynciterator.ts
```diff
@@ -1526,101 +1536,94 @@ export class UnionIterator<T> extends BufferedIterator<T> {
...
     }
-  }
 
-  // Loads sources passed as an iterator
-  protected _loadSources() {
-    // Obtain sources iterator
-    const sources = this._pending!.sources!;
-    delete this._pending!.sources;
+    // @ts-ignore
+    this._sources = (Array.isArray(sources) ? fromArray(sources) : wrap(sources)).map<AsyncIterator<T>>((it: any) => isPromise(it) ? wrap(it as any) : (it as AsyncIterator<T>));
```
jacoscaz, line 1553: I think this could also be simplified after #63

asynciterator.ts
```diff
@@ -1510,12 +1519,13 @@ export class MultiTransformIterator<S, D = S> extends TransformIterator<S, D> {
...
-  @extends module:asynciterator.BufferedIterator
+  @extends module:asynciterator.AsyncIterator
 */
-export class UnionIterator<T> extends BufferedIterator<T> {
-  private _sources : InternalSource<T>[] = [];
-  private _pending? : { sources?: AsyncIterator<MaybePromise<AsyncIterator<T>>> };
-  private _currentSource = -1;
+export class UnionIterator<T> extends AsyncIterator<T> {
+  private _sources : AsyncIterator<AsyncIterator<T>>;
+  private buffer = new LinkedList<AsyncIterator<T>>();
```
jacoscaz, line 1526: I think this needs a `_` prefix.

linkedlist.ts
```diff
@@ -34,8 +34,44 @@ export default class LinkedList<V> {
     return value;
   }
 
+  mutateFilter(filter: (item: V) => boolean) {
+    let last: LinkedNode<V> | null;
+    let next: LinkedNode<V> | null;
+    while (this._head !== null && !filter(this._head.value)) {
```
jacoscaz, line 40: I suspect this could be simplified into one single loop if we were to have the list itself implement the `LinkedNode` interface, referencing the first node in the chain as `next` instead of `_head`. It might not be worth the effort, though.

jacoscaz (commented):
Nicely done, I like how the `maxBufferSize` is used to limit the number of source the iterator reads from. These are just minor nitpicks.

jeswr:
Thanks for the review - I'll wait for #63 to be merged and the rebase and clean this up :)

asynciterator.ts
```diff
@@ -1510,12 +1519,13 @@ export class MultiTransformIterator<S, D = S> extends TransformIterator<S, D> {
...
 /**
   An iterator that generates items by reading from multiple other iterators.
-  @extends module:asynciterator.BufferedIterator
+  @extends module:asynciterator.AsyncIterator
 */
-export class UnionIterator<T> extends BufferedIterator<T> {
-  private _sources : InternalSource<T>[] = [];
-  private _pending? : { sources?: AsyncIterator<MaybePromise<AsyncIterator<T>>> };
-  private _currentSource = -1;
+export class UnionIterator<T> extends AsyncIterator<T> {
```
RubenVerborgh, line 1524: Document whether or not it is in-order; crucial to `prepend` and `append`.

jeswr:
Did work getting this up to date on the train. Will push changes later today

PR #68, opened 2022-04-28
feat: maybeIterator

### LINKED ISSUE #66: AsyncIterator-ext Package
In order to support some more niche use-cases I am coming across I was wondering if it is worth creating an `AsyncIterator-ext` package (see example below). 

IMO the best way to go about this would be:
 - Create an `AsyncIterator` organisation
 - Move the `AsyncIterator` repo to a repo in this organisation **[Requires @RubenVerborgh due to repo permissions]**
 - Create a monorepo of extension components that get released under the `@asynciterator/extension-name` schema (I am happy to take responsibility for maintenance of this new repo)
 - Additionally export all of the extensions in the package `@asynciterator/ext`
 
 The downside of having these in a separate package is that it is hard to use these functions as methods of an iterator (as if we just extended the existing `AsyncIterator` with an `AsyncIteratorExt` class then using a method like `map` would return an `AsyncIterator` class without any of the extension functionality).
 
---

An example of an export from that package would be the following `maybeIterator` function which returns undefined on emptyIterators and an iterator otherwise (the use case for me is to terminate a forward chaining procedure in reasoning components):

```ts
import { AsyncIterator } from 'asynciterator';

/**
 * @param source An AsyncIterator
 * @returns The AsyncIterator if it is not empty, otherwise undefined
 */
async function maybeIterator<T>(source: AsyncIterator<T>): Promise<null | AsyncIterator<T>> {
     // Avoid creating a new iterator where possible
  if ((source instanceof ArrayIterator || source instanceof BufferedIterator) && source._buffer.length > 0) {
     return source
  }
  if (source instanceof IntegerIterator) {
     return source.step >= 0 ? source.next > source.last : source.next : source.last
  }

  let item;
  do {
    if ((item = source.read()) !== null)
      return source.append([item]);
    await awaitReadable(source);
  } while (!source.done);
  return null;
}

function awaitReadable<T>(source: AsyncIterator<T>): Promise<void> {
  return new Promise<void>((res, rej) => {
    if (source.readable || source.done)
      res();

    function done() {
      cleanup();
      res();
    }

    function err() {
      cleanup();
      rej();
    }

    function cleanup() {
      source.removeListener('readable', done);
      source.removeListener('end', done);
      source.removeListener('error', err);
    }

    source.on('readable', done);
    source.on('end', done);
    source.on('error', err);
  });
}
```

jeswr:
Resolves #66

This PR introduces a `maybeIterator` which takes an iterator as input, and returns an iterator *only if* it is non-empty. Otherwise null is returned.

Happy to accept naming suggestions on this one.

Pinging @jacosaz

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -1966,3 +1966,54 @@ type SourceExpression<T> =
 
 type InternalSource<T> =
   AsyncIterator<T> & { _destination: AsyncIterator<any> };
+
+
+/**
+ * @param source An AsyncIterator
+ * @returns The AsyncIterator if it is not empty, otherwise undefined
+ */
+ export async function maybeIterator<T>(source: AsyncIterator<T>): Promise<null | AsyncIterator<T>> {
+  // Avoid creating a new iterator where possible
+  // if ((source instanceof ArrayIterator || source instanceof BufferedIterator) && (source as any)._buffer.length > 0) {
+  //    return source
+  // }
+  // if (source instanceof IntegerIterator && (source as any).step >= 0 ? (source as any).next > (source as any).last : (source as any).next < (source as any).last) {
+  //    return source;
+  // }
+
+  let item;
+  do {
+    if ((item = source.read()) !== null)
+      return source.prepend([item]);
+    await awaitReadable(source);
+  } while (!source.done);
+  return null;
+}
+
+function awaitReadable<T>(source: AsyncIterator<T>): Promise<void> {
+  return new Promise<void>((res, rej) => {
+    if (source.readable || source.done)
+      res();
+
+    function done() {
+      cleanup();
+      res();
+    }
+
+    function err() {
+      cleanup();
+      rej();
+    }
+
+    function cleanup() {
+      source.removeListener('readable', done);
+      source.removeListener('end', done);
+      source.removeListener('error', err);
+    }
+
+    source.on('readable', done);
+    source.on('end', done);
+    source.on('error', err);
+  });
+}
+
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -1967,16 +1967,15 @@ type SourceExpression<T> =
 type InternalSource<T> =
   AsyncIterator<T> & { _destination: AsyncIterator<any> };
 
-
 /**
  * @param source An AsyncIterator
  * @returns The AsyncIterator if it is not empty, otherwise undefined
  */
 export async function maybeIterator<T>(source: AsyncIterator<T>): Promise<null | AsyncIterator<T>> {
   // Avoid creating a new iterator where possible
-  if ((source instanceof ArrayIterator || source instanceof BufferedIterator) && (source as any)._buffer?.length > 0) {
-     return source
-  }
+  if ((source instanceof ArrayIterator || source instanceof BufferedIterator) && (source as any)._buffer?.length > 0)
+    return source;
+
   if (source instanceof IntegerIterator && (source as any)._step >= 0 ? (source as any)._next <= (source as any)._last : (source as any)._next >= (source as any)._last)
     return source;
 
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -623,7 +623,6 @@ export class SingletonIterator<T> extends AsyncIterator<T> {
   }
 }
 
-
 /**
   An iterator that emits the items of a given array.
   @extends module:asynciterator.AsyncIterator
@@ -1971,7 +1970,7 @@ type InternalSource<T> =
  * @param source An AsyncIterator
  * @returns The AsyncIterator if it is not empty, otherwise undefined
  */
-export async function maybeIterator<T>(source: AsyncIterator<T>): Promise<null | AsyncIterator<T>> {
+export async function ensureNonEmpty<T>(source: AsyncIterator<T>): Promise<null | AsyncIterator<T>> {
   return new Promise((res, rej) => {
     let item;
 
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -1972,31 +1972,28 @@ type InternalSource<T> =
  * @returns The AsyncIterator if it is not empty, otherwise undefined
  */
 export async function maybeIterator<T>(source: AsyncIterator<T>): Promise<null | AsyncIterator<T>> {
-  // Avoid creating a new iterator where possible
-  if ((source instanceof ArrayIterator || source instanceof BufferedIterator) && (source as any)._buffer?.length > 0)
-    return source;
+  return new Promise((res, rej) => {
+    let item;
 
-  if (source instanceof IntegerIterator && (source as any)._step >= 0 ? (source as any)._next <= (source as any)._last : (source as any)._next >= (source as any)._last)
-    return source;
+    if ((item = source.read()) !== null) {
+      res(source.prepend([item]));
+      return;
+    }
+    if (source.done) {
+      res(null);
+      return;
+    }
 
-
-  let item;
-  do {
-    if ((item = source.read()) !== null)
-      return source.prepend([item]);
-    await awaitReadable(source);
-  } while (!source.done);
-  return null;
-}
-
-function awaitReadable<T>(source: AsyncIterator<T>): Promise<void> {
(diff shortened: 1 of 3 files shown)


RubenVerborgh:
Could you elaborate on use cases?

jeswr:
The use case for me is to terminate forwards chaining reasoning [here is *some* of the code for it, very WIP](https://github.com/comunica/comunica-feature-reasoning/blob/d89d93b92826058f16ae353ba305580c75aa2a71/packages/actor-rdf-reason-forward-chaining/lib/ActorRdfReasonForwardChaining.ts#L46-L70). Below is a minimal naive example to give you an idea of what the primary logic is

```ts
function getConsequents(quad: RDF.Quad): AsyncIterator<Quad> {
  // Gets any new consequences that can result from adding 'quad' to the dataset with the current rules
}

// New quad to be added to the dataset
let iterator = single(quad);
// N3 Store
const store = new Store();

while ((iterator = maybeIterator(iterator)) !== null) {
  // Get any new quads, by applying rules to quads generated in previous 'loop'
  iterator = union(iterator.map( q => getConsequents(q) ))
  // Add new quads to the store, and filter out from the iterator if already present in the store
  iterator = iterator.filter(q => store.addQuad(q))
}
```

Generally, it can be used to terminate any recursive operation that eventually stops generating new items. 

I recognise that this  is _not_ as not as important as the performance issues that are open, so those should still be addressed before this.

RubenVerborgh:
The `while` loop is a good example!

> ```ts
> while ((iterator = maybeIterator(iterator)) !== null) {
> ```

I take it this is

> ```ts
> while ((iterator = await maybeIterator(iterator)) !== null) {
> ```

but yes, I see the case now. Thinking perhaps to call it `ensureNonEmpty`.

Although for this case (and if this is the general pattern), it could also be

```javascript
do  {
  // Get any new quads, by applying rules to quads generated in previous 'loop'
  iterator = union(iterator.map( q => getConsequents(q) ))
  // Add new quads to the store, and filter out from the iterator if already present in the store
  iterator = iterator.filter(q => store.addQuad(q))
} while (await hasItems(iterator))
```

with

```javascript
function hasItems(iterator) {
  return new Promise((resolve, reject) => {
    iterator.once('error', reject);
    iterator.once('data', () => resolve(true));
    iterator.once('end', () => resolve(false));
  });
}
```

jeswr:
> `await maybeIterator(iterator)`

Ah yes, good catch, and indeed do-while is a better choice a lot of the time. 

> hasItems

The current `hasItems` implementation will result in one element being 'discarded' each time because of the `once` call, which is why there is need to return a new iterator with the item re-appended. Though the logic for retrieving the first element, and then re-appending it may well be improved.

I'm also hesitant about the use of `once` because the `Readable` API will leave an iterator in flowing mode once a when a data listener is unsubscribed (for backwards compat reasons) and I had made changes to align with that in https://github.com/RubenVerborgh/AsyncIterator/pull/43 - this means that an iterator would be left in flowing mode after one element has been emitted using `once`.

In addition I noticed some buggy behavior on a somewhat similar approach (due to the implementation of `toArray`) with
```ts
export async function maybeIterator<T>(source: AsyncIterator<T>): Promise<null | AsyncIterator<T>> {
  const elem = await source.take(1).toArray();
  return elem.length === 1 ? source.prepend(elem) : null;
}
```

where I get the following failing tests

```
 1) maybeIterator
       Should return an iterator with all elements if the iterator is not empty
         fromArray:

      AssertionError: expected [ 1 ] to deeply equal [ 1, 2, 3 ]
      + expected - actual

       [
         1
      +  2
      +  3
       ]
      
      at Context.<anonymous> (file:///home/jeswr/comunica-perf/AsyncIterator/test/maybeIterator-test.js:94:83)

  2) maybeIterator
       Should return an iterator with all elements if the iterator is not empty
         range 1-3:

      AssertionError: expected [ 1 ] to deeply equal [ 1, 2, 3 ]
      + expected - actual

       [
         1
      +  2
      +  3
       ]
      
      at Context.<anonymous> (file:///home/jeswr/comunica-perf/AsyncIterator/test/maybeIterator-test.js:97:74)

  3) maybeIterator
       Should return an iterator with all elements if the iterator is not empty
         MyItemBufferingIterator:

      AssertionError: expected [ 8 ] to deeply equal [ 8, 6, 4, 2, 0 ]
      + expected - actual

       [
         8
      +  6
      +  4
      +  2
      +  0
       ]
      
      at Context.<anonymous> (file:///home/jeswr/comunica-perf/AsyncIterator/test/maybeIterator-test.js:103:92)
```

RubenVerborgh:
> The current `hasItems` implementation will result in one element being 'discarded' each time because of the `once` call,

No, because it's _after_ the filter has been applied? So it's just a second listener?

> I'm also hesitant about the use of once 

Yes, the better implementation is to remove the listeners yourself.

jeswr:
> No, because it's after the filter has been applied? So it's just a second listener?

But won't it still pull the first element out of the list before `iterator = union(iterator.map( q => getConsequents(q) ))` is applied in the next iteration of the `do-while` loop?

> Yes, the better implementation is to remove the listeners yourself.

Even then the readable API leaves the iterator in flowing mode; it requires that you explicitly call `iterator.pause()` in order to stop flowing mode IIRC.

---

https://nodejs.org/api/stream.html#two-reading-modes

For backward compatibility reasons, removing ['data'](https://nodejs.org/api/stream.html#event-data) event handlers will not automatically pause the stream. Also, if there are piped destinations, then calling [stream.pause()](https://nodejs.org/api/stream.html#readablepause) will not guarantee that the stream will remain paused once those destinations drain and ask for more data.

RubenVerborgh:
> But won't it still pull the first element out of the list before `iterator = union(iterator.map( q => getConsequents(q) ))` is applied in the next iteration of the `do-while` loop?

Oof, you're right.

asynciterator.ts
```diff
@@ -1966,3 +1966,53 @@ type SourceExpression<T> =
 
 type InternalSource<T> =
   AsyncIterator<T> & { _destination: AsyncIterator<any> };
+
+/**
+ * @param source An AsyncIterator
+ * @returns The AsyncIterator if it is not empty, otherwise undefined
+ */
+export async function maybeIterator<T>(source: AsyncIterator<T>): Promise<null | AsyncIterator<T>> {
+  // Avoid creating a new iterator where possible
```
RubenVerborgh: I wouldn't bother; you never know what tricks derived classes do.

asynciterator.ts
```diff
@@ -1966,3 +1966,53 @@ type SourceExpression<T> =
...
+    return source;
+
+  if (source instanceof IntegerIterator && (source as any)._step >= 0 ? (source as any)._next <= (source as any)._last : (source as any)._next >= (source as any)._last)
+    return source;
+
+
+  let item;
+  do {
+    if ((item = source.read()) !== null)
+      return source.prepend([item]);
```
RubenVerborgh: I would return `new AsyncIterator` whose `read` method has `if (item) { const rest = item; item = null; return item;} return iterator.read();`.

asynciterator.ts
```diff
@@ -1966,3 +1966,53 @@ type SourceExpression<T> =
...
+    await awaitReadable(source);
```
RubenVerborgh: Yes, but the reattaching might get expensive. I would inline it here so setup and cleanup happen once.

RubenVerborgh (commented):
Some quick thoughts.

asynciterator.ts
jeswr: Once https://github.com/RubenVerborgh/AsyncIterator/pull/65 is resolved/merged I imagine this doesn't produce much of an optimization, so not sure if it is worthwhile doing this.

RubenVerborgh: Yeah; might still be a good local implementation for the quick one-element case.

jeswr: > local implementation for the quick one-element case.

In going to do this I realised that the logic `readable` events should be roughly identical to in https://github.com/RubenVerborgh/AsyncIterator/pull/59 so will wait until that is stable / merged and then we can use the same logic or refactor the necessary logic into a common abstract class.

PR #71, opened 2022-06-24
Fix UnionIterator not destroying sources on close

rubensworks:
This makes the UnionIterator aligned with the default behaviour of TransformIterators.

Given the corresponding behaviour in TransformIterator, I consider this a bugfix, so it probably doesn't require a breaking change.

This was discovered in comunica/comunica#950


RubenVerborgh:
Yes, thanks a lot!

PR #75, opened 2022-06-26
Draft: Feat/mapping iterator

jeswr:
A fresh branch, that does not use transduction, as discussed in #59.

I think this resolves all readability problems @RubenVerborgh though I'm hesitant to say that for certain until I look with fresh eyes in the morning.

~Getting close (I think) to correct readable behaviour - I just need to get one more set of new tests that I just created working. The currently failing tests test the case where the upstream iterator becomes non-readable without closing.~

~I'll come back to this in a few hrs later this evening :)~

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -25,6 +25,11 @@ export function setTaskScheduler(scheduler: TaskScheduler): void {
   taskScheduler = scheduler;
 }
 
+// Returns a function that calls `fn` with `self` as `this` pointer. */
+function bind<T extends Function>(fn: T, self?: object): T {
+  return self ? fn.bind(self) : fn;
+}
+
 /**
   ID of the INIT state.
   An iterator is initializing if it is preparing main item generation.
@@ -161,7 +166,7 @@ export class AsyncIterator<T> extends EventEmitter {
     @param {object?} self The `this` pointer for the callback
   */
   forEach(callback: (item: T) => void, self?: object) {
-    this.on('data', self ? callback.bind(self) : callback);
+    this.on('data', bind(callback, self));
   }
 
   /**
@@ -455,8 +460,8 @@ export class AsyncIterator<T> extends EventEmitter {
     @param {object?} self The `this` pointer for the mapping function
     @returns {module:asynciterator.AsyncIterator} A new iterator that maps the items from this iterator
   */
-  map<D>(map: (item: T) => D, self?: any): AsyncIterator<D> {
-    return this.transform({ map: self ? map.bind(self) : map });
+  map<D>(map: MapFunction<T, D>, self?: any): AsyncIterator<D> {
+    return new MappingIterator(this, bind(map, self));
   }
 
   /**
@@ -469,7 +474,9 @@ export class AsyncIterator<T> extends EventEmitter {
   filter<K extends T>(filter: (item: T) => item is K, self?: any): AsyncIterator<K>;
   filter(filter: (item: T) => boolean, self?: any): AsyncIterator<T>;
   filter(filter: (item: T) => boolean, self?: any): AsyncIterator<T> {
-    return this.transform({ filter: self ? filter.bind(self) : filter });
+    return this.map(function (this: any, item: T) {
+      return filter.call(self || this, item) ? item : null;
+    });
   }
 
   /**
@@ -510,7 +517,7 @@ export class AsyncIterator<T> extends EventEmitter {
     @returns {module:asynciterator.AsyncIterator} A new iterator that skips the given number of items
   */
   skip(offset: number): AsyncIterator<T> {
-    return this.transform({ offset });
+    return this.map(item => offset-- > 0 ? null : item);
   }
 
   /**
@@ -777,9 +784,99 @@ export class IntegerIterator extends AsyncIterator<number> {
   }
 }
 
+/**
+ * A synchronous mapping function from one element to another.
+ * A return value of `null` means that nothing should be emitted for a particular item.
+ */
+export type MapFunction<S, D = S> = (item: S) => D | null;
+
+/** Function that maps an element to itself. */
+export function identity<S>(item: S): typeof item {
+  return item;
+}
+
 
 /**
-  A iterator that maintains an internal buffer of items.
+ An iterator that synchronously transforms every item from its source
+ by applying a mapping function.
+ @extends module:asynciterator.AsyncIterator
+*/
+export class MappingIterator<S, D = S> extends AsyncIterator<D> {
+  protected readonly _map: MapFunction<S, D>;
+  protected readonly _source: InternalSource<S>;
+  protected readonly _destroySource: boolean;
+
+  /**
+   * Applies the given mapping to the source iterator.
+   */
+  constructor(
+    source: AsyncIterator<S>,
+    map: MapFunction<S, D> = identity as MapFunction<S, D>,
+    options: SourcedIteratorOptions = {}
+  ) {
+    super();
+    this._map = map;
+    this._source = ensureSourceAvailable(source);
+    this._destroySource = options.destroySource !== false;
+
+    // Close if the source is already empty
+    if (source.done) {
+      this.close();
+    }
+    // Otherwise, wire up the source for reading
+    else {
+      this._source._destination = this;
+      this._source.on('end', destinationClose);
+      this._source.on('error', destinationEmitError);
+      this._source.on('readable', destinationSetReadable);
+      this.readable = this._source.readable;
+    }
+  }
+
+  /* Tries to read the next item from the iterator. */
+  read(): D | null {
+    if (!this.done) {
+      // Try to read an item that maps to a non-null value
+      if (this._source.readable) {
+        let item: S | null, mapped: D | null;
+        while ((item = this._source.read()) !== null) {
+          if ((mapped = this._map(item)) !== null)
+            return mapped;
+        }
+      }
+      this.readable = false;
+
+      // Close this iterator if the source is empty
+      if (this._source.done)
+        this.close();
+    }
+    return null;
+  }
+
+  /* Cleans up the source iterator and ends. */
+  protected _end(destroy: boolean) {
+    this._source.removeListener('end', destinationClose);
+    this._source.removeListener('error', destinationEmitError);
+    this._source.removeListener('readable', destinationSetReadable);
+    delete this._source._destination;
+    if (this._destroySource)
+      this._source.destroy();
+    super._end(destroy);
+  }
+}
+
+// Validates an AsyncIterator for use as a source within another AsyncIterator
+function ensureSourceAvailable<S>(source?: AsyncIterator<S>, allowDestination = false) {
+  if (!source || !isFunction(source.read) || !isFunction(source.on))
+    throw new Error(`Invalid source: ${source}`);
+  if (!allowDestination && (source as any)._destination)
+    throw new Error('The source already has a destination');
+  return source as InternalSource<S>;
+}
+
+
+/**
+  An iterator that maintains an internal buffer of items.
   This class serves as a base class for other iterators
   with a typically complex item generation process.
   @extends module:asynciterator.AsyncIterator
@@ -797,7 +894,7 @@ export class BufferedIterator<T> extends AsyncIterator<T> {
     @param {integer} [options.maxBufferSize=4] The number of items to preload in the internal buffer
     @param {boolean} [options.autoStart=true] Whether buffering starts directly after construction
   */
-  constructor({ maxBufferSize = 4, autoStart = true } = {}) {
+  constructor({ maxBufferSize = 4, autoStart = true }: BufferedIteratorOptions = {}) {
     super(INIT);
     this.maxBufferSize = maxBufferSize;
     taskScheduler(() => this._init(autoStart));
@@ -1150,14 +1247,10 @@ export class TransformIterator<S, D = S> extends BufferedIterator<D> {
     @param {object} source The source to validate
     @param {boolean} allowDestination Whether the source can already have a destination
   */
-  protected _validateSource(source?: AsyncIterator<S>, allowDestination = false) {
+  protected _validateSource(source?: AsyncIterator<S>, allowDestination = false): InternalSource<S> {
     if (this._source || typeof this._createSource !== 'undefined')
       throw new Error('The source cannot be changed after it has been set');
-    if (!source || !isFunction(source.read) || !isFunction(source.on))
-      throw new Error(`Invalid source: ${source}`);
-    if (!allowDestination && (source as any)._destination)
-      throw new Error('The source already has a destination');
-    return source as InternalSource<S>;
+    return ensureSourceAvailable(source, allowDestination);
   }
 
   /**
@@ -1240,8 +1333,14 @@ export class TransformIterator<S, D = S> extends BufferedIterator<D> {
   }
 }
 
+function destinationSetReadable<S>(this: InternalSource<S>) {
+  this._destination!.readable = true;
+}
 function destinationEmitError<S>(this: InternalSource<S>, error: Error) {
   this._destination!.emit('error', error);
+}
+function destinationClose<S>(this: InternalSource<S>) {
+  this._destination!.close();
 }
 function destinationCloseWhenDone<S>(this: InternalSource<S>) {
   (this._destination as any)._closeWhenDone();
@@ -1956,15 +2055,18 @@ function isSourceExpression<T>(object: any): object is SourceExpression<T> {
   return object && (isEventEmitter(object) || isPromise(object) || isFunction(object));
 }
 
+export interface SourcedIteratorOptions {
+  destroySource?: boolean;
+}
+
 export interface BufferedIteratorOptions {
   maxBufferSize?: number;
   autoStart?: boolean;
 }
 
-export interface TransformIteratorOptions<S> extends BufferedIteratorOptions {
+export interface TransformIteratorOptions<S> extends SourcedIteratorOptions, BufferedIteratorOptions {
   source?: SourceExpression<S>;
   optional?: boolean;
-  destroySource?: boolean;
 }
 
 export interface TransformOptions<S, D> extends TransformIteratorOptions<S> {
--- a/perf/MappingIterator-perf.js
+++ b/perf/MappingIterator-perf.js
@@ -1,0 +1,60 @@
+import { ArrayIterator, range } from '../dist/asynciterator.js';
+
+function noop() {
+  // empty function to drain an iterator
+}
+
+async function perf(warmupIterator, iterator, description) {
+  return new Promise(res => {
+    const now = performance.now();
+    iterator.on('data', noop);
+    iterator.on('end', () => {
+      console.log(description, performance.now() - now);
+      res();
+    });
+  });
+}
+
+function run(iterator) {
+  return new Promise(res => {
+    iterator.on('data', noop);
+    iterator.on('end', () => {
+      res();
+    });
+  });
+}
+
+function baseIterator() {
+  return new ArrayIterator(new Array(20_000_000).fill(true).map((_, i) => i));
+}
+
+function createMapped(filter) {
+  let iterator = baseIterator();
+  for (let j = 0; j < 20; j++) {
+    iterator = iterator.map(item => item);
+    if (filter)
+      iterator = iterator.filter(item => item % (j + 2) === 0);
+  }
+  return iterator;
+}
+
+(async () => {
+  await run(baseIterator()); // warm-up run
+
+  await perf(baseIterator(), createMapped(), '20,000,000 elems 20 maps\t\t\t\t\t');
+  await perf(createMapped(true), createMapped(true), '20,000,000 elems 20 maps 20 filter\t\t\t');
+
+  const now = performance.now();
+  for (let j = 0; j < 100_000; j++) {
+    let it = range(1, 100);
+    for (let k = 0; k < 5; k++)
+      it = it.map(item => item);
+
+    await new Promise((resolve, reject) => {
+      it.on('data', () => null);
+      it.on('end', resolve);
+      it.on('error', reject);
+    });
+  }
+  console.log('100,000 iterators each with 5 maps and 100 elements\t', performance.now() - now);
+})();
--- a/test/MappingIterator-test.js
+++ b/test/MappingIterator-test.js
@@ -1,0 +1,987 @@
+import {
+  AsyncIterator,
+  ArrayIterator,
+  MappingIterator,
+  range,
+  fromArray,
+  wrap,
+  ENDED,
+} from '../dist/asynciterator.js';
+
+import { EventEmitter } from 'events';
+import { expect } from 'chai';
+
+describe('MappingIterator', () => {
+  describe('The MappingIterator function', () => {
+    describe('the result when called with `new`', () => {
+      let instance;
+
+      before(() => {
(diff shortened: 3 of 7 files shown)


jeswr:
Just added a quick performance check we have

Main:
20000000 elems 20 maps 49566.92061899975
20000000 elems 20 maps 20 filters 24873.365212999284
100_000 iterators each with 5 maps and 100 elements  8609.967392999679

This PR
20000000 elems 20 maps 3584.7544269990176
20000000 elems 20 maps 20 filters 3637.598729001358
100_000 iterators each with 5 maps and 100 elements 1951.0139819998294

asynciterator.ts
```diff
@@ -531,7 +536,7 @@ export class AsyncIterator<T> extends EventEmitter {
     @returns {module:asynciterator.AsyncIterator} A new iterator with items in the given range
   */
   range(start: number, end: number): AsyncIterator<T> {
-    return this.transform({ offset: start, limit: Math.max(end - start + 1, 0) });
+    return this.skip(start).take(Math.max(end - start + 1, 0));
```
RubenVerborgh: (I wonder if we want to leave this until we have a better `take`, but not important.)

asynciterator.ts
```diff
@@ -777,9 +789,95 @@ export class IntegerIterator extends AsyncIterator<number> {
   }
 }
 
+/**
+ * A synchronous mapping function from one element to another.
+ * The iterator performing the mapping is passed as a second argument.
```
```suggestion
 * A return value of `null` means that nothing should be emitted for a particular item.
```

asynciterator.ts
```diff
@@ -777,9 +789,95 @@ export class IntegerIterator extends AsyncIterator<number> {
...
+    else {
+      this._source._destination = this;
+      this._source.on('end', destinationClose);
+      this._source.on('error', destinationEmitError);
+      this._source.on('readable', destinationSetReadable);
+      // If we are given a source that is already readable
+      // then we need to set the state of this iterable to readable
+      // also as there is no guarantee that the is no forthcoming
+      // readable event from the source
+      this.readable = this._source.readable;
```
RubenVerborgh, line 832: Makes sense; can also set it to `true` should this cause trouble (it shouldn't).

asynciterator.ts
```diff
@@ -777,9 +789,95 @@ export class IntegerIterator extends AsyncIterator<number> {
...
+
+  read(): D | null {
+    // Do not read the source if the current iterator is ended
+    if (this.done)
+      return null;
+
+    // Do not read from the source if it is ended
+    // TODO: Possibly remove this after https://github.com/RubenVerborgh/AsyncIterator/pull/45 is merged
+    // the requirement for this to be removed is that the 'end' emit occurs synchronously when the ENDED
+    // state is set, which does not currently occur if endAsync is called.
```
RubenVerborgh: I'd probably still keep this in any case. We should not make too strong guarantees about the source being well-behaved.

asynciterator.ts
```diff
@@ -777,9 +789,95 @@ export class IntegerIterator extends AsyncIterator<number> {
...
+    if (this._source.done) {
+      this.close();
+      return null;
+    }
+
+    let item: D | S | null;
```
RubenVerborgh: I'd be conservative here and have `item: S | null` and `mapped: D | null`. Maybe even declare `const mapped` inside of the loop.

asynciterator.ts
```diff
@@ -777,9 +789,95 @@ export class IntegerIterator extends AsyncIterator<number> {
...
+    while ((item = this._source.read()) !== null) {
+      if ((item = this._map(item)) !== null)
+        return item;
+    }
+
+    // This will set readable to false on the current iterator
+    // if the source is no longer readable
+    this.readable = this._source.readable;
```
RubenVerborgh: Right. That's sensible actually.
I'd have been okay with next iteration of `read` triggering that result, but this optimization makes sense. In fact, if the source returns `null`, we _know_ that it is not readable so we can just set it to `false` here?

asynciterator.ts
```diff
@@ -777,9 +789,95 @@ export class IntegerIterator extends AsyncIterator<number> {
...
+      return null;
+
+    // Do not read from the source if it is ended
+    // TODO: Possibly remove this after https://github.com/RubenVerborgh/AsyncIterator/pull/45 is merged
+    // the requirement for this to be removed is that the 'end' emit occurs synchronously when the ENDED
+    // state is set, which does not currently occur if endAsync is called.
+    if (this._source.done) {
+      this.close();
+      return null;
+    }
```
```suggestion
    if (!this._source.readable) {
      this.readable = false;
      if (this._source.done)
        this.close();
      return null;
    }
```

^ We should be maximally well-behaved and not attempt to call `.read` on the source unless it advertises itself as potentially readable.

asynciterator.ts
```diff
@@ -777,9 +789,95 @@ export class IntegerIterator extends AsyncIterator<number> {
...
+    // This will set readable to false on the current iterator
+    // if the source is no longer readable
+    this.readable = this._source.readable;
+    return null;
+  }
+
+  public close() {
+    if (this._destroySource)
+      this._source.destroy();
+    super.close();
```
RubenVerborgh: I think the order should be reversed (first bring ourselves in a closed state, then our internal source). But cross-check with other classes.

asynciterator.ts
```diff
@@ -777,9 +789,95 @@ export class IntegerIterator extends AsyncIterator<number> {
...
+  }
+
+  /* Cleans up the source iterator and ends. */
+  protected _end(destroy: boolean) {
+    this._source.removeListener('end', destinationClose);
+    this._source.removeListener('error', destinationEmitError);
+    this._source.removeListener('readable', destinationSetReadable);
+    delete this._source._destination;
+    if (this._destroySource)
+      this._source.destroy();
```
RubenVerborgh, line 863: That's duplicated from `close`; can it be removed there then (and thus, can the entire `close` implementation be removed)?

perf/MappingIterator-perf.js
```diff
@@ -0,0 +1,46 @@
+import { ArrayIterator, range } from '../dist/asynciterator.js'
+
+for (let i = 0; i < 100_000_000; i++) {
+  // Warming up
```
RubenVerborgh: Nothing's warming though 😅

perf/MappingIterator-perf.js
```diff
@@ -0,0 +1,46 @@
...
+}
+
+function perf(iterator, description) {
+  return new Promise((res) => {
+    for (let i = 0; i < 10_000_000; i++) {
+      // Warming up
```
RubenVerborgh: Still no heat.

RubenVerborgh (commented):
Yes, good stuff, @jeswr. We're absolutely on the right track.

In general, we need to be conservative/defensive when programming derived classes, and not assume too many behaviors of them. I should probably have a list of _allowed assumptions_, but it's not black and white. Briefly, let's not assume that `source` is not one of the implementations we've written, but rather a more unstable one. We have to assume its state and its `readable` are correct, but that's about it.

asynciterator.ts
jeswr: @RubenVerborgh Is it worth opening an issue to apply this idea to other iterators, for instance in [`_readAndTransform`](https://github.com/RubenVerborgh/AsyncIterator/blob/a536bca2dfe84697e1a095c10eb1fcce40f7182c/asynciterator.ts#L1185) and [`_readAndTransformSimple`](https://github.com/RubenVerborgh/AsyncIterator/blob/a536bca2dfe84697e1a095c10eb1fcce40f7182c/asynciterator.ts#L1330-L1339).

asynciterator.ts
jeswr: Setting it to `true` is fine in terms of correctness. In terms of performance, I think it is best to not be readable if the source is not readable as this avoids firing unnecessary `readable` events and in turn prevents any downstream iterators trying to read from this and getting `null` returned.

I think this may also allow a more marked perf. improvement when https://github.com/RubenVerborgh/AsyncIterator/pull/45 is re-addressed as that PR gets rid of a lot of unnecessary `read` calls and in many cases can just about halve the number of `.read()` called when taking unions over transformations.

asynciterator.ts
jeswr: > In fact, if the source returns null, we know that it is not readable

I'm cautious about making that assumption, especially since it doesn't really yield that much in terms of performance gains. For instance, what if someone chose to implement a `FilterIterator` where the `.read` method was

```
read() {
  / * ... readability checks */

  const item = this.source.read()
  return this.filter(item) ? item : null;
}
```
then the source would still be readable.

jeswr: Moreover I can't find anything that makes that assumption explicitly in the [readable API](https://nodejs.org/api/stream.html#event-readable). It makes the statement that "If no data is available to be read, null is returned", but does not say that a `null` return implies it is not readable.

asynciterator.ts
jeswr: Ah yes it can, the only thing that changes is that the destruction happens a tick later.

This will become moot with https://github.com/RubenVerborgh/AsyncIterator/pull/45 anyway which removes the `endAsync` that results in this next-tick behavior.

asynciterator.ts
jacoscaz: I wouldn’t be surprised if a filter method like that were to subtly break things in some use cases. I do think there’s an implicit assumption that a stream that reads null is also a stream that is not readable at that point in time. A safer implementation would read until the source reads null or until the source reads an item that matches the filter (as per what happens in `MappingIterator`).

jacoscaz: Also, consider this paragraph from https://nodejs.org/api/stream.html#readablereadsize :

> When reading a large file `.read()` may return `null`, having consumed all buffered content so far, but there is still more data to come not yet buffered. In this case a new 'readable' event will be emitted when there is more data in the buffer.

AFAIU, this implies that whenever we `read()` the value `null` we can stop reading and resume doing so upon a new `readable` event.

asynciterator.ts
jeswr: I've made this change. The only thing I should flag is that in https://github.com/RubenVerborgh/AsyncIterator/pull/75/commits/b34fb6e71647fd6c58afe0734271864d6220461a I had to change from waiting a tick (where tests were failing) to doing a `beforeEach` which waits for the end event on the iterator that is being closed, and then will wait another tick before testing for destruction.

I think this should be ok, but is probably something that we can make even safer in #45 as noted in the comments in that PR.

asynciterator.ts
jeswr: I've been convinced, change pushed

asynciterator.ts
RubenVerborgh: Note that `TransformIterator` and `SimpleTransformIterator` also accept regular Node.js `Readable` instances, which do not define `readable`. So the goal there is to also work with non-`AsyncIterator`s (see `wrap`).

We can have the same goal for `MappingIterator`, but that requires different tests.

RubenVerborgh:
Thanks, @jeswr, looking really good. Amazing what two fresh pairs of eyes can do, innit? 😉
I will make some small stylistic changes the code here and there, and then assign back to you for final approval.
Then we merge and release a new minor (unless you want to have `take` / the `UntilIterator` in as well).

jeswr:
I think that since this is close, clean it up and then we can release. After that is done we could put most of the logic into an abstract class that `MapIterator`, `UntilIterator` and, eventually `CompositeIterator` can share.

I have in mind that it will look something like this (which does have all tests current passing and is over 2x faster on the current performance tests):

```ts
/**
 * A synchronous mapping function from one element to another.
 * A return value of `null` means that nothing should be emitted for a particular item.
 */
export type MapFunction<S, D = S> = (item: S) => D | null;

/**
 An iterator that calls a synchronous operation
 on every item from its source iterator.
 @extends module:asynciterator.AsyncIterator
*/
export abstract class SynchronousTransformIterator<S, D = S> extends AsyncIterator<D> {
  private readonly _destroySource: boolean;
  protected _source: InternalSource<S>;

  /**
   * Applies the given mapping to the source iterator.
   */
  constructor(
    _source: AsyncIterator<S>,
    options: SourcedIteratorOptions = {}
  ) {
    super();
    this._source = _validateSource(_source);
    this._destroySource = options.destroySource !== false;

    if (_source.done) {
      this.close();
    }
    else {
      this._source._destination = this;
      this._source.on('end', destinationClose);
      this._source.on('error', destinationEmitError);
      this._source.on('readable', destinationSetReadable);
      // If we are given a source that is already readable
      // then we need to set the state of this iterable to readable
      // also as there is no guarantee that the is no forthcoming
      // readable event from the source
      this.readable = this._source.readable;
    }
  }

  protected abstract safeRead(): D | null;

  read(): D | null {
    // Do not read the source if the current iterator is ended
    if (this.done)
      return null;

    // A source should only be read from if readable is true
    if (!this._source.readable) {
      this.readable = false;
      if (this._source.done)
        this.close();
      return null;
    }

    const item = this.safeRead();

    if (item !== null) {
      return item;
    }

    // This will set readable to false on the current iterator
    // if the source is no longer readable
    this.readable = false;
    return null;
  }

  /* Cleans up the source iterator and ends. */
  protected _end(destroy: boolean) {
    this._source.removeListener('end', destinationClose);
    this._source.removeListener('error', destinationEmitError);
    this._source.removeListener('readable', destinationSetReadable);
    delete this._source._destination;
    if (this._destroySource)
      this._source.destroy();
    super._end(destroy);
  }
}

export class MappingIterator<S, D = S> extends SynchronousTransformIterator<S, D> {
  constructor(
    source: AsyncIterator<S>,
    private readonly _map: MapFunction<S, D>,
    options: SourcedIteratorOptions = {}
  ) {
    super(source, options);
  }

  safeRead() {
    let item: S | null;
    while ((item = this._source.read()) !== null) {
      const mapped = this._map(item);
      if (mapped !== null)
        return mapped;
    }
    return null;
  }

  map<K>(map: MapFunction<D, K>, self?: any): AsyncIterator<K> {
    return new CompositeMappingIterator(this._source, [this._map, bind(map, self)], this);
  }
}

export class CompositeMappingIterator<S, D = S> extends SynchronousTransformIterator<S, D> {
  constructor(
    private root: AsyncIterator<S>,
    private mappings: MapFunction<any, any>[] = [],
    source: AsyncIterator<any>,
    options: SourcedIteratorOptions = {},
  ) {
    super(source, options);
  }

  safeRead() {
    // TODO: See if this is actually necessary
    // A source should only be read from if readable is true
    if (!this.root.readable) {
      this.readable = false;
      // TODO: See if this should be here
      if (this.root.done)
        this.close();
      return null;
    }

    let mapped : any = null;
    while (mapped === null && (mapped = this.root.read()) !== null) {
      for (let i = 0; i < this.mappings.length; i++) {
        mapped = this.mappings[i](mapped);
        if (mapped === null)
          break;
      }
    }
    return mapped;
  }

  map<K>(map: MapFunction<D, K>, self?: any): AsyncIterator<K> {
    return new CompositeMappingIterator(this.root, [...this.mappings, bind(map, self)], this);
  }
}

export class UntilIterator<S> extends SynchronousTransformIterator<S> {
  constructor(
    source: AsyncIterator<S>,
    private limit: number,
    options: SourcedIteratorOptions = {}
  ) {
    super(source, options);
  }

  safeRead() {
    if (this.limit <= 0) {
      this.close();
      return null;
    }
    this.limit -= 1;
    return this._source.read();
  }
}
```

I'll open up a branch with a cleaned up, and more heavily tested version of the code above (minus the `CompositeMappingIterator`) once the current branch is merged. The `CompositeMappingIterator` can then be cleanly addressed later on.

test/MappingIterator-test.js
```diff
@@ -0,0 +1,1026 @@
...
+import { EventEmitter } from 'events';
+import { expect } from 'chai';
+
+describe('MappingIterator', () => {
+  describe('The MappingIterator function', () => {
+    describe('the result when called with `new`', () => {
+      let instance;
+
+      before(() => {
+        // TODO: Re-enable this
```
RubenVerborgh, line 21: @jeswr What's this one?

RubenVerborgh:
@jeswr I have done a review and have arrived at the code you can now see in this PR.
Apart from an overall approval, 3 questions:
- Any concerns about the simplified `read` logic? https://github.com/RubenVerborgh/AsyncIterator/pull/75/commits/eeb47967333b8116c3124f1857c9cf677a9d6ad4#diff-8c1065e179bf8c72bbaa6ae735f3be78aac1e609456c0086be7d2b3c1dae66ecR844
- Any concerns about the`readable` logic? https://github.com/RubenVerborgh/AsyncIterator/pull/75/commits/eeb47967333b8116c3124f1857c9cf677a9d6ad4#diff-8c1065e179bf8c72bbaa6ae735f3be78aac1e609456c0086be7d2b3c1dae66ecL1334-L1337 The original was a bit naughty as superfluous events would be emitted; I'm just checking if there was a reason for that. The unit tests worked fine without it.
- Any concerns related to the changes and your proposed `SynchronousTransformIterator`?

jeswr:
> Any concerns about the simplified read logic?

 - I usually call `this._source.read()` outside of `if` statements as this reduces the amount of branching required to get to the call which I assume to be the most common case when `.read()` is called. ~Would need to do perf. checks to see if this *actually* has any performance impact though.~ Looks like this doesn't make a difference in practise though

```
Original

[jeswr@orangejuly AsyncIterator]$ node perf/MappingIterator-perf.js 
20000000 elems 20 maps                                   1678.1333249993622
20000000 elems 20 maps 20 filter                         1854.2675620019436
100_000 iterators each with 5 maps and 100 elements      2297.956739999354
[jeswr@orangejuly AsyncIterator]$ node perf/MappingIterator-perf.js 
20000000 elems 20 maps                                   1611.3272369988263
20000000 elems 20 maps 20 filter                         1734.021374002099
100_000 iterators each with 5 maps and 100 elements      2286.2911899983883
[jeswr@orangejuly AsyncIterator]$ node perf/MappingIterator-perf.js 
20000000 elems 20 maps                                   1594.8407809995115
20000000 elems 20 maps 20 filter                         1748.050572000444
100_000 iterators each with 5 maps and 100 elements      2127.236763998866

---

Now

[jeswr@orangejuly AsyncIterator]$ node perf/MappingIterator-perf.js 
20000000 elems 20 maps                                   1703.3545530028641
20000000 elems 20 maps 20 filter                         1865.0185250006616
100_000 iterators each with 5 maps and 100 elements      2192.7923329994082
[jeswr@orangejuly AsyncIterator]$ node perf/MappingIterator-perf.js 
20000000 elems 20 maps                                   1602.2818830013275
20000000 elems 20 maps 20 filter                         1791.3512960001826
100_000 iterators each with 5 maps and 100 elements      2325.8517739996314
[jeswr@orangejuly AsyncIterator]$ node perf/MappingIterator-perf.js 
20000000 elems 20 maps                                   1611.5303689986467
20000000 elems 20 maps 20 filter                         1763.5889509990811
100_000 iterators each with 5 maps and 100 elements      2306.6571919992566

```
 
 - In terms of getting the most out of #45 once it is merged, I think the `.done` check should be an `if` afterwards rather than an `else if` https://github.com/RubenVerborgh/AsyncIterator/blob/eeb47967333b8116c3124f1857c9cf677a9d6ad4/asynciterator.ts#L854-L856 to allow chained iterators to close faster.

jeswr:
> Any concerns about thereadable logic? https://github.com/RubenVerborgh/AsyncIterator/commit/eeb47967333b8116c3124f1857c9cf677a9d6ad4#diff-8c1065e179bf8c72bbaa6ae735f3be78aac1e609456c0086be7d2b3c1dae66ecL1334-L1337 The original was a bit naughty as superfluous events would be emitted; I'm just checking if there was a reason for that. The unit tests worked fine without it.

Ooh interesting - I thought those were failing before without. I think it was historical when @jacoscaz needed a similar pattern to get unit tests passing.

If tests are working then I think it is safe to get rid of it provided we have good enough coverage with asynchronous `source`s for the iterator.

jeswr:
> Any concerns related to the changes and your proposed SynchronousTransformIterator?

Nope - tests still pass making the changes to convert to the `SynchronousTransformIterator` proposal.

~I don't think so - it should just be a matter of moving the `.safeRead()` call. I'll quickly run some tests to confirm.~

asynciterator.ts
```diff
@@ -777,9 +791,88 @@ export class IntegerIterator extends AsyncIterator<number> {
...
+    if (!this.done) {
+      if (this._source.readable) {
+        // Try to read an item that maps to a non-null value
+        let item: S | null, mapped: D | null;
+        while ((item = this._source.read()) !== null) {
+          if ((mapped = this._map(item)) !== null)
+            return mapped;
+        }
+      }
+      else if (this._source.done) {
```
jeswr:
```suggestion
      if (this._source.done) {
```

Only one suggested change - see https://github.com/RubenVerborgh/AsyncIterator/pull/75#issuecomment-1173590862.

jeswr (commented):
"Approved" @RubenVerborgh  with one suggested change (I don't have permissions to click the actual review button :))

RubenVerborgh:
> I usually call `this._source.read()` outside of `if` statements as this reduces the amount of branching

Yeah; and I tend to optimize for fewer `return` points. The performance should not be impacted (and thanks for confirming that).

> Ooh interesting - I thought those were failing before without.
> If tests are working then I think it is safe to get rid of it provided we have good enough coverage with asynchronous sources for the iterator.

Yes—and that's the benefit of not messing with the base `readable` mechanism. It contains several guards and is ridiculously well tested for all kinds of orders. So we use the same logic now and inherit the same protections.

> "Approved" @RubenVerborgh with one suggested change (I don't have permissions to click the actual review button :))

And fully agreed—the code is even leaner now. I really like what we landed on:

```typescript
  read(): D | null {
    if (!this.done) {
      // Try to read an item that maps to a non-null value
      if (this._source.readable) {
        let item: S | null, mapped: D | null;
        while ((item = this._source.read()) !== null) {
          if ((mapped = this._map(item)) !== null)
            return mapped;
        }
      }
      this.readable = false;

      // Close this iterator if the source is empty
      if (this._source.done)
        this.close();
    }
    return null;
  }
```

This has been a common occurrence for me with `AsyncIterator`: lots of trying and fiddling, and eventually, you end up with really simple code. However, all of the fiddling resulted in lots of tests for edge cases that you wouldn't have found even if you would have thought of the simple code at the start (and you'd probably have gotten 1 or 2 lines wrong, but wouldn't know).

asynciterator.ts
```diff
@@ -469,7 +474,9 @@ export class AsyncIterator<T> extends EventEmitter {
   filter<K extends T>(filter: (item: T) => item is K, self?: any): AsyncIterator<K>;
   filter(filter: (item: T) => boolean, self?: any): AsyncIterator<T>;
   filter(filter: (item: T) => boolean, self?: any): AsyncIterator<T> {
-    return this.transform({ filter: self ? filter.bind(self) : filter });
+    return this.map(function (this: any, item: T) {
+      return filter.call(self || this, item) ? item : null;
+    });
```
RubenVerborgh, line 479: @jeswr Found this nasty bug still when restoring all of the main tests. (The `this` pointer `filter` is called with would be wrong because of the nested function.)

RubenVerborgh:
Thanks a lot, @jeswr, excellent work. Thanks @jacoscaz as well.

jacoscaz:
I had very little to do with this! Great job. I'll rebase #63 ASAP.

RubenVerborgh:
@jacoscaz Reviews and chats are important! See Gitter 😉

rubensworks:
Nice work!

@jeswr Do you have any insights into the performance impact on Comunica?

PR #76, opened 2022-06-27
Fix MultiTransformIterator not destroying transformers on close

rubensworks:
This was discovered in comunica/comunica#950

This may also be the root-cause of comunica/comunica#989


RubenVerborgh:
Quite the oversight 😱 thanks!!

PR #77, opened 2022-07-04
draft: synchronousTransformIterator

jeswr:
As discussed in https://github.com/RubenVerborgh/AsyncIterator/pull/75#issuecomment-1168219877.

TODOs
 - Add more coverage
 - See if https://github.com/RubenVerborgh/AsyncIterator/blob/3a5b62cf8ca00e60013b6aa7c1e692b26fcb6964/asynciterator.ts#L905-L911 is necessary. @RubenVerborgh the question here is - can we just assume the `root` is `.readable` if the parent source is `.readable`. If we assume that all of the intermediary sources are `SynchronousTransformIterator`s as is the case when doing chaining then this will be true; but I am not sure if we should be making this assumption.
 
Quick testing indicates that this results in a ~2x performance increase on the tests in #75

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -527,7 +527,7 @@ export class AsyncIterator<T> extends EventEmitter {
     @returns {module:asynciterator.AsyncIterator} A new iterator with at most the given number of items
   */
   take(limit: number): AsyncIterator<T> {
-    return this.transform({ limit });
+    return new UntilIterator(this, limit);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


jacoscaz:
Should this be a part of the changes summarized at  https://github.com/RubenVerborgh/AsyncIterator/issues/44#issuecomment-1166907641 ?

jacoscaz:
@jeswr is this something you're still working on? Would this be releasable as 3.x or 4.x?

jeswr:
@jacoscaz The changes are still valid - and there are no breaking changes. I can try and work on the test coverage over the weekend depending on my time (though I'm looking a bit stretched thin at this point).

jacoscaz:
No pressure or rush meant! Just checking, I was having a look at updating my summary in #44 and wasn't sure about this one. Nice!

PR #78, opened 2022-07-30
Symbol for _destination property

jacoscaz:
This PR complements #63 (which should be merged first) by renaming the `_destination` property to a dedicated `Symbol`, guaranteed to be unique, so that we can never trigger property collisions when setting `_destination` on externally-provided source objects for which `isAsyncReadable(source) === true` at https://github.com/RubenVerborgh/AsyncIterator/blob/159ed2b09e599d03d574aec2c98b40ae416d168b/asynciterator.ts#L2030-L2043 .

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -791,6 +791,9 @@ export function identity<S>(item: S): typeof item {
   return item;
 }
 
+/** Key indicating the current consumer of a source. */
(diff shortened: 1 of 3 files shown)


RubenVerborgh:
Thanks, @jacoscaz, I'm happy to take it from here!

PR #79, opened 2022-08-01
V3 unioniterator

jeswr:
supercedes #65

The bolded performance regression needs to be ironed out first.

Before
For loop with 5x10^9 elems: 5.207s
UnionIterator 2x10^7 iterators: 30.297s
**UnionIterator 1000x500 iterators: 1.897s**
UnionIterator 1000x500 iterators - max parallelism of 1: 1.837s

After
For loop with 5x10^9 elems: 6.349s
UnionIterator 2x10^7 iterators: 3.513s
**UnionIterator 1000x500 iterators: 7.219s**
UnionIterator 1000x500 iterators - max parallelism of 1: 168.007ms

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -1653,7 +1653,7 @@ export class UnionIterator<T> extends AsyncIterator<T> {
     if (this._sources.done && this._sourceStarted)
       this.close();
     else
-      this.readable = true;
+      this.readable = this._sources.readable;
   }
 
   // Adds the given source to the internal sources array
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -481,8 +481,8 @@ export class AsyncIterator<T> extends EventEmitter {
     @param {Array|module:asynciterator.AsyncIterator} items Items to insert before this iterator's (remaining) items
     @returns {module:asynciterator.AsyncIterator} A new iterator that prepends items to this iterator
   */
-  prepend(items: T[] | AsyncIterator<T>): AsyncIterator<T> {
-    return this.transform({ prepend: items });
+  prepend(items: IterableSource<T>): AsyncIterator<T> {
+    return union([wrap(items), this]);
   }
 
   /**
@@ -491,8 +491,8 @@ export class AsyncIterator<T> extends EventEmitter {
     @param {Array|module:asynciterator.AsyncIterator} items Items to insert after this iterator's (remaining) items
     @returns {module:asynciterator.AsyncIterator} A new iterator that appends items to this iterator
   */
-  append(items: T[] | AsyncIterator<T>): AsyncIterator<T> {
-    return this.transform({ append: items });
+  append(items: IterableSource<T>): AsyncIterator<T> {
+    return union([this, wrap(items)]);
   }
 
   /**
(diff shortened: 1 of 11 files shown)


asynciterator.ts
```diff
@@ -1621,145 +1621,124 @@ export class MultiTransformIterator<S, D = S> extends TransformIterator<S, D> {
...
-    super._end(destroy);
-
+  public close() {
+    this._removeSource(this._sources);
     // Destroy all sources that are still readable
     if (this._destroySources) {
-      for (const source of this._sources)
+      for (const source of this._buffer) {
+        this._removeSource(source);
         source.destroy();
```
jeswr: This one of 2 behaviors I am not happy about here is that we don't also destroy any sources that have not been produced by the `this._sources` iterator. But I can't think of a clean way of doing this here as `this._sources` may have a large amount of overhead to generate the rest of the sources to the end

asynciterator.ts
```diff
@@ -1621,145 +1621,124 @@ export class MultiTransformIterator<S, D = S> extends TransformIterator<S, D> {
...
 
-  // Loads pending sources into the sources list
-  protected _loadSources() {
-    // Obtain sources iterator
-    const sources = this._pending!.sources!;
-    this._pending!.loading = true;
+  // Adds the given source to the internal sources array
+  protected _addSource(source: InternalSource<any>) {
+    source[DESTINATION] = this;
+    source.on('error', destinationEmitError);
```
jeswr, line 1669: This is the second of two behaviors I am not happy with. In particular; errors will only be propogated once we have pulled it from `this._source`. An error may occur before this.

test/UnionIterator-test.js
```diff
@@ -226,22 +239,9 @@ describe('UnionIterator', () => {
...
       });
-
-      it('should pass errors', () => {
-        const callback = sinon.spy();
-        const error = new Error('error');
-        iterator.once('error', callback);
-        sourceIterator.emit('error', error);
-        callback.should.have.been.calledOnce;
-        callback.should.have.been.calledWith(error);
-      });
```
jeswr, line 244: This is an example of an [error that can no longer be propogated](https://github.com/RubenVerborgh/AsyncIterator/pull/79#discussion_r934928512) with the changes we have made.

test/UnionIterator-test.js
```diff
@@ -580,9 +547,18 @@ describe('UnionIterator', () => {
...
+      iterator.read();
+      iterator.close();
+
+      await new Promise(resolve => iterator.on('end', resolve));
+
+      sourcesIterator.closed.should.be.true;
 
       sources[0].closed.should.be.true;
-      sources[1].closed.should.be.true;
+      sources[1].closed.should.be.false;
```
jeswr, line 561: This is an example of [where a source could not be destroyed](https://github.com/RubenVerborgh/AsyncIterator/pull/79#discussion_r934927797) and hence closed because it had not yet been pulled from `this._source`.

RubenVerborgh:
See https://github.com/RubenVerborgh/AsyncIterator/pull/81

PR #80, opened 2022-08-04
fix: re-enable IterableSource options in wrap

jeswr:
Prior to this fix uses of `wrap` following the pattern

```ts
const iterator = new EventEmitter();
wrap(iterator, { autoStart: false })
```

were breaking due to the fact that the type overloads did not include the optional option

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -2089,7 +2089,8 @@ export class WrappingIterator<T> extends AsyncIterator<T> {
   @param {object} [options] Settings of the iterator
   @returns {module:asynciterator.AsyncIterator} A new iterator with the items from the given iterator
 */
-export function wrap<T>(source: null | undefined): AsyncIterator<T>;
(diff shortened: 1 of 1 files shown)


jacoscaz (approved):
Looks good to me!

RubenVerborgh:
Thanks for finding this!
Isn't just removing the overloads altogether the easier option? 25085c4a5b2f5fa5f15979cc8f9266e42465cfa3

jeswr:
> Isn't just removing the overloads altogether the easier option?

Seems like the sensible option to me @RubenVerborgh - could you release the current state of the main branch as 3.6.1 unless @jacoscaz has any objections?

jacoscaz:
No objections!

RubenVerborgh:
Thanks both! https://www.npmjs.com/package/asynciterator/v/3.6.1

PR #82, opened 2022-08-11
feat: export LinkedList

jeswr:
Pros: Makes this accessible to downstream applications (I want to use it in a custom iterator downstream)
Cons: Will require a major version bump if we want to remove the LinkedList class or otherwise make breaking changes to its API

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -4,11 +4,14 @@
  */
 
 import { EventEmitter } from 'events';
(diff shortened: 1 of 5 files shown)


jeswr:
bump @RubenVerborgh

RubenVerborgh:
Good call, I think a linked list will be a relevant data structure for consumers in any case.

Issue #83, opened 2022-08-17
WrappingIterator does not destroy sources


rubensworks:
Unlike other iterators, the new WrappingIterator does not destroy sources on end, nor has it a `destroySource` option.

This will likely result in problems where iterators keep running internally for a while (or infinitely), even though they can be stopped.

jacoscaz:
Good point. Is anyone already working on this? I can have a look if not.

rubensworks:
I'm not working on it, so feel free to have a go at it @jacoscaz :-)

For this, it would be good to also keep #84 in mind.

PR #84, opened 2022-08-17
Fix slow async sources not being destroyed

rubensworks:
When a TransformIterator receives an async callback as input, but the callback is very slow, race conditions could occur when destroying the iterator.
Concretely, when the iterator was closed/destroy before the callback resolves, the underlying iterator would never be destroyed.

FYI, this caused major problems in the context with link traversal, where the query would end, but underlying link traversal iterators (which can be infinite) kept on running.

This fix should also be taken into account when #83 is being fixed.

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -1206,16 +1206,17 @@ export class TransformIterator<S, D = S> extends BufferedIterator<D> {
   }
 
   set source(value: AsyncIterator<S> | undefined) {
-    // Do not change sources if the iterator is already done
-    if (this.done)
-      return;
-
     // Validate and set source
     const source = this._source = this._validateSource(value);
     source[DESTINATION] = this;
 
-    // Close this iterator if the source has already ended
-    if (source.done) {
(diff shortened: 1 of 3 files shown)


RubenVerborgh:
Oops that's very worrying. Thanks, let me have a look.

RubenVerborgh:
Note for v4.0: we should avoid needing to update this in multiple places.

RubenVerborgh:
@rubensworks Counter-proposal at 87006a5ad23ad77ec860bb0142dfe5c623aeb043

The problem was broader: other aspects of setting a source on a closed iterator were also inconsistent with the behavior on open iterators (such as checking whether the source is valid).

rubensworks:
@RubenVerborgh Looks good to me! Just tested it on my end, and everything still seems to work with the new fix.

RubenVerborgh:
Great, published as v3.6.2!

rubensworks:
Thanks!

PR #85, opened 2022-08-17
Adds destroySource support to wrap() and WrappingIterator

### LINKED ISSUE #83: WrappingIterator does not destroy sources

jacoscaz:
Closes #83

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -1998,9 +1998,11 @@ class HistoryReader<T> {
  */
 export class WrappingIterator<T> extends AsyncIterator<T> {
   protected _source: InternalSource<T> | null = null;
+  protected _destroySource: boolean;
 
-  constructor(source?: MaybePromise<IterableSource<T>>) {
+  constructor(source?: MaybePromise<IterableSource<T>>, opts?: SourcedIteratorOptions) {
     super();
+    this._destroySource = opts?.destroySource !== false;
 
     // If promise, set up a temporary source and replace when ready
     if (isPromise(source)) {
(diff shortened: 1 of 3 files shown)


rubensworks (approved):
LGTM!

Just noting that since WrappingIterator doesn't support `autoStart: false`, it won't be picked up yet by Comunica.

jacoscaz:
@rubensworks I guess `WrappingIterator` implicitly behaves as if `autoStart` is always set to `false`. Perhaps we can make it so that if only `autoStart: false` is passed we can return a `WrappingIterator` instead of a `TransformIterator`?

rubensworks:
> I guess WrappingIterator implicitly behaves as if autoStart is always set to false

Not sure, at first glance it looks to me like `source` is immediately being set, and things like listeners are being attached.
I suspect some more changes would have to be made, similar to `TransformIterator`.

jacoscaz:
@rubensworks I think I've done it in https://github.com/RubenVerborgh/AsyncIterator/pull/85/commits/80fb650d7b06274ed80c61e0194a11a95b2621e7 but we can rollback if it doesn't look right.

jacoscaz:
Mh... Nope, nevermind. I tested it with one project before pushing but it did break another. I'll rollback and consider this PR complete. We can address this separately.

jeswr:
>  I'll rollback and consider this PR complete. We can address this separately.

bump @RubenVerborgh ?

RubenVerborgh:
Thanks @jacoscaz, great work!

PR #91, opened 2022-09-21
Allow wrapping arrays with options

### LINKED ISSUE #88: potential bug: `wrap` now autostarting iterators that were previously not autostarted prior to optimisation work
Placeholder - will try create a minimal example or debug tommorow :)

RubenVerborgh:
Fixes https://github.com/RubenVerborgh/AsyncIterator/issues/88

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -2123,9 +2123,12 @@ export class WrappingIterator<T> extends AsyncIterator<T> {
 */
 export function wrap<T>(source?: MaybePromise<IterableSource<T>> | null,
                         options?: TransformIteratorOptions<T>): AsyncIterator<T> {
-  // TransformIterator if TransformIteratorOptions were specified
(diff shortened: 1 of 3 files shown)


jacoscaz:
Yup, +1 from me!

jeswr:
@RubenVerborgh - I also think w need to be closing the wrapping iterator here https://github.com/RubenVerborgh/AsyncIterator/blob/915a85734824809cc65b189d3e095c590da1334b/asynciterator.ts#L2031 (otherwise the iterator is permanently open if the promise rejects).

RubenVerborgh:
@jeswr We probably need a better strategy around `error`. At the moment, I consider an iterator that has emitted an `error` as unpredictable and unusable. Perhaps we need a separated `errored` state. I'd leave that as a todo in https://github.com/RubenVerborgh/AsyncIterator/issues/92

jeswr (approved):

RubenVerborgh:
Thank you!

PR #94, opened 2022-11-08
Use .js extension for CJS, and .mjs for ESM

### LINKED ISSUE #1097: process is not defined using comunica and react without editing the webpack
Hello I'm trying to use comunica with a standard react app (created with for example ***npx create-reactapp my-app***) without editing the webpack (when for example ***npm run eject*** is used to edit the webpack).

Comunica works perfect with react when you edit the webpack and use "node-polyfil-webpack-plugin". But gives the following error when a standard react app is used:
image

I'm looking at this problem for a while but can't really figure out how to fix it. I'm currently comparing the 2 app's with each other.
Both use the newest version of comunica. So again, I want to change the code of comunica in order to work, but not that of the react app itself. 

I already scoured the internet for this problem, but none really worked. 
I've tried to add the following code in App.js: 
`    window.process = { `
`      env: {`
`          NODE_ENV: 'development'`
`      }`
`    }    `
which solves the issue but gives a new error: ***process.tick is not defined***, so it must be solved in another way.
Any help or tips would be welcome!

rubensworks:
This fixes issues with the commonly-used create-react-app tool that is unable to handle .cjs files.

This change consider CJS by default for .js files, but still exposes ESM for the tools that have proper support for it.

See facebook/create-react-app#11889

Closes comunica/comunica#1097

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,16 +10,16 @@
   ],
   "license": "MIT",
   "repository": "RubenVerborgh/AsyncIterator",
-  "type": "module",
-  "main": "dist/asynciterator.cjs",
+  "type": "commonjs",
+  "main": "dist/asynciterator.js",
   "exports": {
     "./package.json": "./package.json",
     ".": {
-      "import": "./dist/asynciterator.js",
-      "require": "./dist/asynciterator.cjs"
+      "require": "./dist/asynciterator.js",
+      "import": "./dist/asynciterator.mjs"
     }
   },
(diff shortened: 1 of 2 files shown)


RubenVerborgh:
Thanks, @rubensworks, this helps a lot. Releasing as next minor version now.

RubenVerborgh:
Thanks, @rubensworks, this helps a lot. Releasing as next minor version now.

PR #99, opened 2023-05-28
Fix ClonedIterators not being readable

rubensworks:
Since ClonedIterators are not autoStarted, they should be marked as readable. Otherwise, dependents may never consider them as being readable.

This lies at the root cause of the following Comunica bugs, where execution would halt due to an intermediary iterator never becoming readable:
- comunica/comunica#1168
- comunica/comunica#817

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -1796,6 +1796,9 @@ export class ClonedIterator<T> extends TransformIterator<T> {
   constructor(source: AsyncIterator<T>) {
     super(source, { autoStart: false });
     this._reading = false;
+    // As cloned iterators are not auto-started, they must always be marked as readable.
+    if (source)
+      this.readable = true;
   }
 
   protected _init() {
--- a/test/ClonedIterator-test.js
+++ b/test/ClonedIterator-test.js
@@ -173,8 +173,8 @@ describe('ClonedIterator', () => {
         });
 
         if (index === 0) {
-          it('should not have emitted the `readable` event', () => {
-            getClone()._eventCounts.readable.should.equal(0);
+          it('should have emitted the `readable` event once', () => {
+            getClone()._eventCounts.readable.should.equal(1);
           });
         }
 
@@ -187,8 +187,8 @@ describe('ClonedIterator', () => {
         });
 
         if (index === 0) {
-          it('should not be readable', () => {
-            getClone().readable.should.be.false;
+          it('should be readable', () => {
+            getClone().readable.should.be.true;
           });
 
           it('should return null on read', () => {
@@ -204,7 +204,7 @@ describe('ClonedIterator', () => {
           before(() => { getIterator()._push('a'); });
 
         it('should have emitted the `readable` event', () => {
-          getClone()._eventCounts.readable.should.equal(1);
+          getClone()._eventCounts.readable.should.equal(index === 0 ? 2 : 1);
         });
 
         it('should not have emitted the `end` event', () => {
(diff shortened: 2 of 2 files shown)


rubensworks:
Not sure about the test failure, as it seems to be happening in code that this PR does not touches, and only for Node 10 (which is out of LTS).

asynciterator.ts
```diff
@@ -1796,6 +1796,9 @@ export class ClonedIterator<T> extends TransformIterator<T> {
   constructor(source: AsyncIterator<T>) {
     super(source, { autoStart: false });
     this._reading = false;
+    // As cloned iterators are not auto-started, they must always be marked as readable.
+    if (source)
```
jeswr, line 1800: I suspect that rather than *always* setting the iterator to be readable if there is a source, it will be sufficient to add `|| source.readable` to the [condition here](https://github.com/RubenVerborgh/AsyncIterator/blob/3171521759194b3ce463f989d8253e35978ef690/asynciterator.ts#L1836-L1837).

jeswr (changes requested):

asynciterator.ts
rubensworks: I just tried, and it looks like this doesn't fix the two Comunica issues mentioned above anymore unfortunately.

jeswr: Hmm ok. I still don't like just always setting `.readable` to be true here because we are glossing over some kind of race condition taking place in the iterator.

Second to that setting an iterator to be readable when it is not has problems for performance because it triggers all downstream iterators to call `.read()` only to get no results.

Is it possible that the problematic cases are occurring when we are calling `.clone()` on a source that has already ended. In that case it might be that we need to be checking `source.done` instead of `source.readable` in the conditional on line 1836.

rubensworks: > Is it possible that the problematic cases are occurring when we are calling .clone() on a source that has already ended. In that case it might be that we need to be checking source.done instead of source.readable in the conditional on line 1836.

Just tried this as well, not solving the problem either unfortunately.


Note that this change is essentially just restoring the behaviour of `BufferedIterator`, which is setting `readable` to `true` if autostart is disabled: https://github.com/RubenVerborgh/AsyncIterator/blob/3171521759194b3ce463f989d8253e35978ef690/asynciterator.ts#L971
We need to do this, since `ClonedIterator` is voiding the implementation of `BufferedIterator`'s `_init`.

jeswr:
Doing a bit of dig into this at the moment; for https://github.com/comunica/comunica/issues/1168 I've observed that the issue is only occuring when taking the `.clone()` of `CloneIterators` in the `INIT` `_state`. Will come back to this later in the day.

*note*: I was expecting https://github.com/RubenVerborgh/AsyncIterator/pull/100/files to be a minimal repro of this but sadly it was not so will continue digging.

rubensworks:
@RubenVerborgh (or @jeswr?) Any chance this PR might be merged?
I'm sure @pheyvaer would be very happy to see this issue resolved.

jeswr (approved):
Sorry - this completely dropped off my radar. In the future feel free to bump me after a couple of days of non-response to issues/PRs :).

Since the concern in this discussion https://github.com/RubenVerborgh/AsyncIterator/pull/99#discussion_r1209581971 is not that this is a buggy PR - but rather that we hide an underlying bug, and possibly introduce a minor performance hit; I am in favor of merging this PR and opening up in issue to investigate as a follow up.

I will leave @RubenVerborgh to merge as I do not have rights to make a subsequent npm release.

RubenVerborgh:
Thanks! The test failure is indeed very weird. Not that we need Node 10 support, but it's sometimes a canary for things going wrong.

RubenVerborgh:
The test failure was just a fluke, works fine on main. Releasing next.

PR #102, opened 2023-07-04
Make iterators AsyncIterable, Closes #89

### LINKED ISSUE #89: Feat: Make asynciterator asynciterable
In particular this logic https://github.com/LDflex/LDflex-Comunica/blob/f01ece6fa8cee720061efd5ddd23ed82921555e4/src/ComunicaEngine.ts#L158-L215 should be implemented as the `[Symbol.asyncIterator]` method of `AsyncIterator`

rubensworks:
While this will likely come at the cost of lower performance compared to plain data-events, we should add this for use cases where code readability is preferred over performance.

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -580,6 +580,11 @@ export class AsyncIterator<T> extends EventEmitter implements AsyncIterable<T> {
       next(): Promise<IteratorResult<T>> {
         return new Promise<IteratorResult<T>>((resolve, reject) => {
           it.on('error', reject);
+          function nextCallback(): void {
+            it.removeListener('readable', nextCallback);
+            it.removeListener('end', nextCallback);
+            tryResolve();
+          }
           function tryResolve(): void {
             const value = it.read();
             if (value !== null) {
@@ -594,11 +599,6 @@ export class AsyncIterator<T> extends EventEmitter implements AsyncIterable<T> {
             }
             else {
               // In all other cases, wait for the iterator to become readable or ended
-              const nextCallback = () => {
-                it.removeListener('readable', nextCallback);
-                it.removeListener('end', nextCallback);
-                tryResolve();
-              };
               it.once('readable', nextCallback);
               it.once('end', nextCallback);
             }
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -579,31 +579,31 @@ export class AsyncIterator<T> extends EventEmitter implements AsyncIterable<T> {
     return {
       next(): Promise<IteratorResult<T>> {
         return new Promise<IteratorResult<T>>((resolve, reject) => {
+          let hasDataListeners = false;
           it.on('error', reject);
-          function nextCallback(): void {
-            it.removeListener('readable', nextCallback);
-            it.removeListener('end', nextCallback);
-            tryResolve();
-          }
-          function tryResolve(): void {
+
+          // Tries to read the next item and passes it through the callback
+          function readNext(): void {
+            // Try to read the next item
             const value = it.read();
-            if (value !== null) {
-              // Immediately resolve if we have read an element
+            if (value !== null || it.done) {
               it.removeListener('error', reject);
-              resolve({ done: false, value });
+              if (hasDataListeners) {
+                it.removeListener('end', readNext);
+                it.removeListener('readable', readNext);
+              }
+              resolve(value !== null ?
+                { done: false, value } :
+                { done: true, value: undefined });
             }
-            else if (it.done) {
-              // Close if we're done
-              it.removeListener('error', reject);
-              resolve({ done: true, value: undefined });
-            }
-            else {
-              // In all other cases, wait for the iterator to become readable or ended
-              it.once('readable', nextCallback);
-              it.once('end', nextCallback);
+            // If none available, listen for new items
+            else if (!hasDataListeners) {
+              hasDataListeners = true;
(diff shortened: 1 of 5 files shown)


asynciterator.ts
```diff
@@ -568,6 +567,47 @@ export class AsyncIterator<T> extends EventEmitter {
...
+              it.removeListener('error', reject);
+              resolve({ done: true, value: undefined });
+            }
+            else {
+              // In all other cases, wait for the iterator to become readable or ended
+              const nextCallback = () => {
+                it.removeListener('readable', nextCallback);
+                it.removeListener('end', nextCallback);
+                tryResolve();
+              };
```
jeswr: You should be able to define this outside of `tryResolve` rather than re-defining it every time.

jeswr (approved):
LGTM (barring the minor optimisation above)

Note that this logic achieves the same effect over in LDflex https://github.com/LDflex/LDflex-Comunica/blob/f01ece6fa8cee720061efd5ddd23ed82921555e4/src/ComunicaEngine.ts#L161-L215.

I will leave @RubenVerborgh to merge as I do not have rights to make a subsequent npm release.

asynciterator.ts
rubensworks: Done!

rubensworks:
@RubenVerborgh Could you have a look at this one? Will certainly make many comunica users very happy :-)

RubenVerborgh:
@rubensworks So I first did f60ceccdc98f87dbafe2ceddcfc84de4e506a4d6 to minimize listener re-attachment within a single `next` iteration, but then I realize we could keep listeners attached in 43cedf106d1a6564682f3249e1a3c01ac867a19a, and then realized we _must_ do that to catch errors in between `next` calls. However, the test coverage is not complete for those cases yet.

rubensworks:
I also had a `pendingError`-like approach in mind before, but I decided to not implement it like that before due to the chance of errors getting lost.
What could happen, is that the AsyncIterable is only being consumed partially, that an error is emitted after the last consumed element, and that the error will get lost because no further elements are getting consumed.
The alternative is that unhandled errors get thrown, but at least the errors do not get lost that way 😅 

So I guess we'll have to choose between either errors to always be explicitly forwarded (but them sometimes getting lost on partial consumption), or errors never getting lost, but them sometimes being unhandled errors.

----

After writing this, I just realized that we could also implement our own unhandled exception approach. If we notice (somehow) that the AsyncIterable is not consumed anymore, but we have a pending error, that we throw an unhandled exception. This might give us the best of both worlds.

RubenVerborgh:
> After writing this, I just realized that we could also implement our own unhandled exception approach. If we notice (somehow) that the AsyncIterable is not consumed anymore, but we have a pending error, that we throw an unhandled exception. This might give us the best of both worlds.

I like the approach, except I'm not sure how to implement it with the ECMAScript `AsyncIterator` contract. How do I _know_ that no `next` will be called, if the error arrives between `next` calls?
What I am missing is a `destroy` method on ECMAScript `AsyncIterator`.

rubensworks:
> How do I know that no next will be called, if the error arrives between next calls?

We may be able to do something by registering the created `AsyncIterator` into a [FinalizationRegistry](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry), but it feels finicky...

RubenVerborgh:
Ooof yes. Or perhaps for now at a comment that we recommend adding an error listener to the iterator if you don't plan on fully consuming it (since you can't detach a `for await` created thing yourself, because you don't have a reference to it).

_Or_ we could keep a list per iterator of the ESAsyncIterator instances created (or even allow only one), and then _destroy_ on the iterator also destroys all of them.

Actually, _destroy_ on the iterator might just work because it would also delete listeners.

rubensworks:
@RubenVerborgh At last, I found some time to look into finalizing the unit tests and documenting the error events edge-case.

RubenVerborgh:
That is quite amazing, thanks. So we good to go? Anything I should still do? (I see a review is pending, just checking if you need specific things.)

rubensworks:
@RubenVerborgh Should be feature-complete AFAICS, so it should be ready to be merged if you agree :-)

asynciterator.ts
```diff
@@ -611,7 +616,7 @@ export class AsyncIterator<T> extends EventEmitter implements AsyncIterable<T> {
         currentResolve = currentReject = pendingError = null;
         removeListeners();
       }
-      else if (pendingError !== null) {
+      else if (pendingError === null) {
```
RubenVerborgh, line 619: 🤦 obvs I put that in on purpose to check your tests

RubenVerborgh:
I was _so_ pleased that my code turned out to be the one in the end, only the see the dreadful `!==` error.
You go ahead and merge it, you did the tests and that's the hard work!

rubensworks:
Merged! :-)

@RubenVerborgh, before you release, it might make sense to include #105 as well.

jacoscaz:
@rubensworks @RubenVerborgh brilliant!!!

rubensworks:
@RubenVerborgh Could you publish a new release? I'd like to include it in the next major release of Comunica, which I intend to release soon.

RubenVerborgh:
@rubensworks Published v3.9.0! Thanks for your patience and continued support.

PR #104, opened 2023-11-24
Fix 3 tests where the wrong event is checked.

maartyman:
I found 3 tests where the wrong event was checked. The title said 'should not have listeners for the `data` event' and in the test the 'readable' event was checked for listeners.


RubenVerborgh:
Thank you, @maartyman!

PR #106, opened 2024-06-18
UnionIterator skips not readable subiterators

maartyman:
The UnionIterator is used in [the bind join in comunica](https://github.com/comunica/comunica/blob/c3dd70a424e175877fd883277e9d0191bc15bf62/packages/actor-rdf-join-inner-multi-bind/lib/ActorRdfJoinMultiBind.ts#L78) (and [incremunica](https://github.com/maartyman/incremunica/blob/28f4c24ca9122f27de6197239e4deb07e786a2b7/packages/actor-rdf-join-inner-incremental-computational-multi-bind/lib/ActorRdfJoinInnerIncrementalComputationalMultiBind.ts#L164)). During some performance tests, I saw that the UnionIterator called `read()` on all its sub-iterators, even if only one was readable. This PR fixes this by adding a check before calling `read()` on the sub-iterators. The sub-iterators need to be read to start them, that is why I added the `read` attribute to the _sources. To be fair, I think there is a better way to check if the sources have started and if not call `read()` on them.

### DIFF
--- a/asynciterator.ts
+++ b/asynciterator.ts
@@ -1717,7 +1717,7 @@ export class MultiTransformIterator<S, D = S> extends TransformIterator<S, D> {
   @extends module:asynciterator.BufferedIterator
 */
 export class UnionIterator<T> extends BufferedIterator<T> {
-  private _sources : InternalSource<T>[] = [];
+  private _sources : {read: boolean, source: InternalSource<T>}[] = [];
   private _pending? : { loading: boolean, sources?: AsyncIterator<MaybePromise<AsyncIterator<T>>> };
   private _currentSource = -1;
   protected _destroySources: boolean;
@@ -1784,10 +1784,14 @@ export class UnionIterator<T> extends BufferedIterator<T> {
     if (isPromise(source))
       source = wrap<T>(source) as any as InternalSource<T>;
     if (!source.done) {
-      this._sources.push(source);
+      const sourceObj = { read: true, source };
+      this._sources.push(sourceObj);
       source[DESTINATION] = this;
       source.on('error', destinationEmitError);
-      source.on('readable', destinationFillBuffer);
+      source.on('readable', () => {
+        sourceObj.read = true;
+        destinationFillBuffer.bind(<InternalSource<unknown>>sourceObj.source)();
+      });
       source.on('end', destinationRemoveEmptySources);
     }
   }
@@ -1796,9 +1800,9 @@ export class UnionIterator<T> extends BufferedIterator<T> {
   protected _removeEmptySources() {
     this._sources = this._sources.filter((source, index) => {
       // Adjust the index of the current source if needed
-      if (source.done && index <= this._currentSource)
+      if (source.source.done && index <= this._currentSource)
         this._currentSource--;
-      return !source.done;
+      return !source.source.done;
     });
     this._fillBuffer();
   }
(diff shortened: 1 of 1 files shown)


asynciterator.ts
```diff
@@ -1717,7 +1717,7 @@ export class MultiTransformIterator<S, D = S> extends TransformIterator<S, D> {
   @extends module:asynciterator.BufferedIterator
 */
 export class UnionIterator<T> extends BufferedIterator<T> {
-  private _sources : InternalSource<T>[] = [];
+  private _sources : {read: boolean, source: InternalSource<T>}[] = [];
```
```suggestion
  private _sources : { read: boolean, source: InternalSource<T> }[] = [];
```
weird that the linter didn't whine. It should. Something wrong with CI?

asynciterator.ts
```diff
@@ -1784,10 +1784,14 @@ export class UnionIterator<T> extends BufferedIterator<T> {
...
       source = wrap<T>(source) as any as InternalSource<T>;
     if (!source.done) {
-      this._sources.push(source);
+      const sourceObj = { read: true, source };
+      this._sources.push(sourceObj);
       source[DESTINATION] = this;
       source.on('error', destinationEmitError);
-      source.on('readable', destinationFillBuffer);
+      source.on('readable', () => {
+        sourceObj.read = true;
```
RubenVerborgh: Let's just use a local variable instead of an object.

But also, what does `read` mean? Is is the past tense, as in, `done`? Or current tense?
I blame English grammar here.

RubenVerborgh (commented):
Thanks @maartyman. I know these are hard to find and solve.
Nonetheless I'll need to ask for a test that breaks without but works with.

That's to avoid accidental breakage in the future.

Code-wise, can we just make the `read` thing a simple variable?

asynciterator.ts
RubenVerborgh: Where are we reading this value though?

maartyman:
I'm sorry, in my hastiness I forgot the skip check in the _read() function.

maartyman:
Also, why is does the UnionIterator extend the BufferedIterator?

RubenVerborgh:
> why does the UnionIterator extend the BufferedIterator?

Short answer: so readers have to wait less (it fills up in the background).

Longer answer: buffering should've been mix-in functionality rather than inherited functionality.

maartyman:
Okay, I'm asking because I see a performance increase if I use a non buffered UnionIterator. But this is just in my tests with querying local stores with incremunica, when using online sources this might be different.

RubenVerborgh:
We can give it a go if the performance effect extends to other cases. An easy fix could be to set the buffer size to zero. Perhaps there's already sufficient intermediary buffering happening in your case.

jeswr:
There is a closed PR https://github.com/RubenVerborgh/AsyncIterator/pull/81 which does it and has significant improvements - it was meant to be bundled into a bugger update that I never got around to.

asynciterator.ts
```diff
@@ -1817,8 +1821,11 @@ export class UnionIterator<T> extends BufferedIterator<T> {
         // Pick the next source
         this._currentSource = (this._currentSource + 1) % this._sources.length;
         const source = this._sources[this._currentSource];
+        if (!source.source.readable && !source.requiresRead)
```
rubensworks, line 1824: Is there a reason reason why only having `!source.source.readable` is insufficient?

maartyman: It's to jumpstart the sub-iterators, I added this because I had issues with it in the past. If I remove it, all test still pass, so I guess it's not necessary? I guess an iterator should be created with readable set to true...

rubensworks: > I guess an iterator should be created with readable set to true...

Yes, for iterators without autoStart, readable should be true.

PR #109, opened 2025-05-30
perf: uses setImmediate() implementation from tiny-set-immediate

jacoscaz:
This PR changes the task scheduler to use an externally-provided macrotask scheduling function and defaults to using the `setImmediate()` implementation from [`tiny-set-immediate`](https://github.com/jquense/tiny-set-immediate).

This is done to work around performance issues discovered in https://github.com/comunica/comunica/discussions/1552 and ultimately traced back to the enormous difference between `setImmediate()`, which is only available is Node.js-like environments, and `setTimeout(fn, 0)`, which is what the in-house `scheduleMacrotask` function would fallback to prior to this change. The former is roughly 90x faster than the latter.

### DIFF
--- a/taskscheduler.ts
+++ b/taskscheduler.ts
@@ -1,14 +1,11 @@
+
 const resolved = Promise.resolve(undefined);
 
 // Returns a function that asynchronously schedules a task
-export function createTaskScheduler() : TaskScheduler {
+export function createTaskScheduler(scheduleMacrotask: TaskScheduler) : TaskScheduler {
   // Use or create a microtask scheduler
   const scheduleMicrotask = typeof queueMicrotask === 'function' ?
     queueMicrotask : (task: Task) => resolved.then(task);
-
-  // Use or create a macrotask scheduler
-  const scheduleMacrotask = typeof setImmediate === 'function' ?
-    setImmediate : (task: Task) => setTimeout(task, 0);
 
   // Interrupt with a macrotask every once in a while to avoid freezing
   let i = 0;
--- a/taskscheduler.ts
+++ b/taskscheduler.ts
@@ -1,4 +1,3 @@
-
 const resolved = Promise.resolve(undefined);
 
 // Returns a function that asynchronously schedules a task
--- a/taskscheduler.ts
+++ b/taskscheduler.ts
@@ -1,7 +1,9 @@
+import { setImmediate } from 'tiny-set-immediate';
+
 const resolved = Promise.resolve(undefined);
 
 // Returns a function that asynchronously schedules a task
-export function createTaskScheduler(scheduleMacrotask: TaskScheduler) : TaskScheduler {
+export function createTaskScheduler(scheduleMacrotask: TaskScheduler = setImmediate) : TaskScheduler {
   // Use or create a microtask scheduler
   const scheduleMicrotask = typeof queueMicrotask === 'function' ?
     queueMicrotask : (task: Task) => resolved.then(task);
--- a/test/TaskScheduler-test.js
+++ b/test/TaskScheduler-test.js
@@ -51,28 +51,27 @@ describe('TaskScheduler', () => {
     });
   });
 
-  describe('a task scheduler when setImmediate exists', () => {
+  describe('a task scheduler using an externally-provided setImmediate() implementation', () => {
     const backups = {};
+    let setImmediate = sinon.spy();
 
     beforeEach(() => {
       backups.setTimeout = global.setTimeout;
       backups.queueMicrotask = global.queueMicrotask;
-      backups.setImmediate = global.setImmediate;
-      global.setImmediate = sinon.spy();
       global.setTimeout = sinon.spy();
       global.queueMicrotask = sinon.spy();
+      setImmediate = sinon.spy();
     });
 
     after(() => {
       global.setTimeout = backups.setTimeout;
       global.queueMicrotask = backups.queueMicrotask;
-      global.setImmediate = backups.setImmediate;
       delete global.window;
     });
 
     it('alternates between setImmediate and queueMicrotask', () => {
       // Perform 100 calls
-      const taskScheduler = createTaskScheduler();
+      const taskScheduler = createTaskScheduler(setImmediate);
       const tasks = [];
       for (let i = 0; i < 100; i++) {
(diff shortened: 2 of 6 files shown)


test/TaskScheduler-test.js
```diff
@@ -51,28 +51,27 @@ describe('TaskScheduler', () => {
     });
   });
 
-  describe('a task scheduler when setImmediate exists', () => {
```
rubensworks, line 54: I wonder if we want to keep the two variants of these tests, just to validate the behaviour of `tiny-set-immediate`?

tsconfig.json
```diff
@@ -39,7 +39,7 @@
     "noFallthroughCasesInSwitch": true,       /* Report errors for fallthrough cases in switch statement. */
 
     /* Module Resolution Options */
-    // "moduleResolution": "node",            /* Specify module resolution strategy: 'node' (Node.js) or 'classic' (TypeScript pre-1.6). */
+    "moduleResolution": "node",               /* Specify module resolution strategy: 'node' (Node.js) or 'classic' (TypeScript pre-1.6). */
```
rubensworks, line 42: This change may have been unintentional?

taskscheduler.ts
```diff
@@ -1,15 +1,12 @@
+
```
rubensworks, line 2: Empty line can be removed.

rubensworks (changes requested):
Are you seeing any (negative) impact when running the perf script?

test/TaskScheduler-test.js
jacoscaz: If that's the case shouldn't we add more tests to `tiny-set-immediate`'s own test suite?

tsconfig.json
jacoscaz: This is intentional as without this change the TS compiler doesn't know how to resolve the `tiny-set-immediate` package name.

taskscheduler.ts
jacoscaz: Done in https://github.com/RubenVerborgh/AsyncIterator/pull/109/commits/ef08da67c146398ee64b49fcd225c892036b8f66

jacoscaz:
> Are you seeing any (negative) impact when running the perf script?

From `main`:

```
20,000,000 elems 20 maps					 3672.3449579999997
20,000,000 elems 20 maps 20 filter			 2590.885416
100,000 iterators each with 5 maps and 100 elements	 2878.7446660000005
```

From this PR:

```
20,000,000 elems 20 maps					 3685.9225420000002
20,000,000 elems 20 maps 20 filter			 2610.698125000001
100,000 iterators each with 5 maps and 100 elements	 2886.2840000000006
```

taskscheduler.ts
```diff
@@ -1,15 +1,11 @@
 const resolved = Promise.resolve(undefined);
 
 // Returns a function that asynchronously schedules a task
-export function createTaskScheduler() : TaskScheduler {
+export function createTaskScheduler(scheduleMacrotask: TaskScheduler) : TaskScheduler {
```
RubenVerborgh: Make it optional argument, defaulting to `tiny-set-immediate`? Then the code from `asynciterator.ts` can stay as-is; it shouldn't need to know about internals.

test/TaskScheduler-test.js
```diff
@@ -51,28 +51,27 @@ describe('TaskScheduler', () => {
...
+  describe('a task scheduler using an externally-provided setImmediate() implementation', () => {
```
RubenVerborgh, line 54:
```suggestion
  describe('a task scheduler with default setImmediate implementation', () => {
```

test/TaskScheduler-test.js
```diff
@@ -51,28 +51,27 @@ describe('TaskScheduler', () => {
...
     const backups = {};
+    let setImmediate = sinon.spy();
```
RubenVerborgh, line 56: just let it default here; or spy on `tiny-set-immediate` itself

test/TaskScheduler-test.js
```diff
@@ -121,48 +120,6 @@ describe('TaskScheduler', () => {
     });
   });
 
-  describe('a task scheduler when setImmediate does not exist', () => {
```
RubenVerborgh, line 124: Maybe keep this test, but repurpose for custom implementation? I like that the test verifies the alternating behavior in another way.

```
describe('a task scheduler with custom setImmediate implementation',
```

test/TaskScheduler-test.js
```diff
@@ -176,7 +133,7 @@ describe('TaskScheduler', () => {
     });
 
     it('can schedule a task', done => {
-      const taskScheduler = createTaskScheduler();
+      const taskScheduler = createTaskScheduler(setImmediate);
```
RubenVerborgh, line 136: suggest to leave unchanged

RubenVerborgh (approved):
Great work!

taskscheduler.ts
jacoscaz: Done in https://github.com/RubenVerborgh/AsyncIterator/pull/109/commits/fc2d24701b737fe9eb31d24d283632c2fa555b8e

jacoscaz:
@RubenVerborgh I couldn't find a way to spy on `tiny-set-immediate`'s exports directly so I explicitly split the tests into two different sub-suites: one for the no-op spy, whose callback must be invoked manually, and one for a spy on `tiny-set-immediate`'s `setImmediate()`, in which the callback is invoked automatically and thus the numerical checks are postponed via `setImmediate()` itself.

jacoscaz:
@rubensworks @RubenVerborgh just checkin in on this PR. I have not addressed all of the comments above as I wasn't able to find a way to spy on `tiny-set-immediate`'s exports, hence the idea of splitting into two sub-suites (see previous comment).

Insofar as I remember this should be good to go, though perhaps it'd be worth adding an explicit mention to `tiny-set-immediate` in `README.md` to inform downstream users about the performance differences and why it would be a good idea to use it.

No pressure, as always. Over the next few weeks I'll probably experiment with using  `asynciterator` for a major refactor and this morning it dawned on me that I hadn't followed up on this.

rubensworks (approved):
Oh, I forgot about this one.
Would be good to get this one merged indeed!

